Issue #86, opened 2021-05-31
Accessing session on the client side receives the session store.


pixelmund:
While implementing 'refetch queries' i noticed that the session passed to fetchQuery is a svelte store on the client side, it works fine in `load` which makes sense.
If you add this log to the example 
```typescript
async function fetchQuery({ text, variables = {} }, session: any) {
	console.log(session)

	const result = await this.fetch('http://' + API_URL, {
		method: 'POST',
		headers: {
			'Content-Type': 'application/json',
		},
		body: JSON.stringify({
			query: text,
			variables,
		}),
	})

	// parse the result as json
	return await result.json()
}
```
You can see that session is `{subscribe: ƒ, set: ƒ, update: ƒ}`

We should probably use `get` from svelte/store in mutations and refetch.

AlecAivazis:
Nice catch - i agree that we probably want to use `get` in those situations

Issue #57, opened 2021-03-05
Refetch queries


AlecAivazis:
Queries sometimes need to be refetched, possibly with new arguments

plunkettscott:
Currently have a use case where I need to manipulate filters on the page and re-run a query with the filters adjusted (from query variables). Is this the use case you are referring to here?

AlecAivazis:
Yep! that's exactly right. Atm, your best bet is to embed as much as you can in the URL filters so you can use page transitions to cause a new query to be fired.

If you feel up for adding this, I don't think many changes are needed. [Here](https://github.com/AlecAivazis/houdini/blob/main/packages/houdini/runtime/query.ts#L64) is where the return value from `query` is defined. I'm pretty sure all it would take would be adding the function to this object that fires the query (using `fetchQuery` from the exported runtime) and passes the result to the same `writeData` function in that object to update the cache values and subscriptions for the query.

pixelmund:
I started working on this [here](https://github.com/pixelmund/houdini/tree/refetch-queries).

I have some issues with the houdini build:cmd:main command `sed -i '#!/usr/bin/env node' build/cmd.js`, sed: undefined label. Im on Mac and it seems like mac needs another required argument on sed -i [https://mpdaugherty.wordpress.com/2010/05/27/difference-with-sed-in-place-editing-on-mac-os-x-vs-linux/](https://mpdaugherty.wordpress.com/2010/05/27/difference-with-sed-in-place-editing-on-mac-os-x-vs-linux/)

Otherwise its working pretty well, can u look at this @AlecAivazis and tell me what, or if im missing something in my implementation.

pixelmund:
One thing that came to my mind is: Should we return the result of the fetch `{data, error}` or just void?

AlecAivazis:
Feel free to remove that whole `sed` command from package.json - its not necessary and is left overs from the kit effort that I have been meaning to remove

plunkettscott:
> One thing that came to my mind is: Should we return the result of the fetch `{data, error}` or just void?

I think returning the `data` object might make sense. In cases where you just want differentiate from the _new_ data, I think it makes sense to capture the return value (i.e., data tables where you show X records per page). Of course, this should still update the cache.

AlecAivazis:
I wonder if differentiating new and old data is better done in a reactive statement so that we don't have to complicate houdini's api 🤔

plunkettscott:
[...]

PR #87, opened 2021-05-31
Refetch queries #57 #86

pixelmund:


packages/houdini/runtime/query.ts
```diff
@@ -61,30 +65,89 @@ export default function query<_Query extends Operation<any, any>>(
...
-		},
+		writeData,
+		refetch,
 	}
 }
 
 // we need to wrap the response from a query in something that we can
 // use as a proxy to the query for refetches, writing to the cache, etc
-type QueryResponse<_Data, _Input> = {
+type QueryResponse<_Data, _Input, _Result> = {
```
AlecAivazis, line 147: Is this third type parameter used anywhere?

packages/houdini/runtime/query.ts
```diff
@@ -61,30 +65,89 @@ export default function query<_Query extends Operation<any, any>>(
...
+		if (subscriptionSpec) {
+			cache.subscribe(subscriptionSpec, variables)
+		}
+
+		// write the data we received
+		cache.write(artifact.selection, newData.data, variables)
+	}
+
+	function refetch(newVariables?: _Query['input']) {
+		return new Promise<any>(async (resolve, reject) => {
```
AlecAivazis, line 83: This function shares a lot of logic with the network bits of `mutation`. I think it would be a good idea to define a function in `network.ts` with this logic to avoid the duplication.

packages/houdini/runtime/query.ts
```diff
@@ -61,30 +65,89 @@ export default function query<_Query extends Operation<any, any>>(
 		)
 	})
 
+	const sessionStore = getSession()
+
+	function writeData(newData: _Query['result'], newVariables: _Query['input']) {
```
AlecAivazis, line 70: Just fyi, you don't need to pull this function out of the object in order to use it in another method. Here's an example: https://www.typescriptlang.org/play?#code/MYewdgzgLgBAZiEMC8MDeBYAUDXMAWApgDbEgAUAlOtnnTFPgJYQB0A7iAE7EAmVtPAF8ANINyce-aphz1coSCGKFWZAObkA5EVIgAhFsriYQ7GazYEIVrrJUgA

I personally prefer to keep them defined in that single object literal in order to keep the block of code where the functions are defined easily referenceable

AlecAivazis (changes requested):
Thanks for getting this out so quickly @pixelmund! I really appreciate all the work you are putting in on this library. Just a few changes but we should be able to get this in quickly

pixelmund:
Good changes! I'll try to get these in tomorrow. My dog is getting put to sleep today and I'm too sad right now.

AlecAivazis:
Oh no!!! That's awful. I'm so sorry. 

No worries at all about the changes - take your time.

pixelmund:
Thank you! At least he got very old(16, cocker spaniel) and had a happy, carefree life. Everyone has to go someday.

packages/houdini/runtime/query.ts
pixelmund: I don't know why but this is not working for me, it is fine in this [playground](https://www.typescriptlang.org/play?alwaysStrict=false&target=99&jsx=0#code/AQ4YwewOwZwF2AMwhYBeYBvAsAKFAcAPRHBwAWApsPBAE7UzkQCuANgCbABG1DAhhwC00NgE8kdCAFsyVcPzZtKdAOQxgABxUxtYOAEsAbpTyFQHfnH4AuLMAC+AGjPngDRJThhyACipKEHbwdAZQAOYAlFiubqCQsBDKAHRsEOG+FAYwTsABaZEA3LFx7l4sdFDAUJQA7sAAClLS2ZS+-DBiUGDAvgwwSSa5DABWlPrRaAB8Mfil5iXzIAkDKWkZWTBFeItLm8m1oXCUACJW-L4AjADM17mXAEzX27i75g4vcc5vxKQsMJQuJpQtJ+KFxDwJBRqMDKMCIGBKDABnQyKgANaUOHANJgRQ0azHYBhGhdHq1AwUH4kOTUSzWYDhYyUKrcCSwtKCH6HSmnc6+Gq1M7WOxQFjSXh0XKCgBqYIM-G4yhgovFkuiODm8xWSUoqXSArqwv40rqctCiuVkR+3y1jh2duQEGSHi8Pl8qlq9E4qkiQA) but when i try to implement it i always get `Uncaught (in promise) TypeError: Cannot read property 'writeData' of undefined`. If i log `this` it is also undefined.
```js
    const foo = {
        // the store should be read-only from the caller's perspective
        data: { subscribe: store.subscribe },
        // used primarily by the preprocessor to keep local state in sync with
        // the data given by preload
        refetch(newVariables) {
            return new Promise(async (resolve, reject) => {
                let result;
                const session = get(sessionStore);
                // since we have a promise that's wrapping async/await we need a giant try/catch that will
                // reject the promise
                try {
                    const fetchCtx = {
                        fetch: window.fetch.bind(window),
                        session,
                        context: {},
                        page: {
                            host: '',
                            path: '',
                            params: {},
                            query: new URLSearchParams(),
                        },
                    };
                    // pull the query text out of the compiled artifact
                    const { raw: text } = artifact;
                    const res = await fetchQuery(fetchCtx, {
                        text,
                        // Pass the new Variables or the initial ones
                        variables: newVariables || variables,
                    }, session);
                    // we could have gotten a null response
                    if (res.errors) {
                        reject(res.errors);
                        return;
                    }
                    if (!res.data) {
                        reject([new Error('Encountered empty data response in query payload')]);
                        return;
                    }
                    result = res;
                }
                catch (e) {
                    reject(e);
                    return;
                }
                this.writeData(result, newVariables);
                resolve(result.data);
            });
        },
        writeData(newData, newVariables) {
            variables = newVariables || {};
            // make sure we list to the new data
            if (subscriptionSpec) {
                cache.subscribe(subscriptionSpec, variables);
            }
            // write the data we received
            cache.write(artifact.selection, newData.data, variables);
        },
    }
    return foo;
```

What am i missing?

pixelmund: I found this one which explains it prelly well i think.
[https://javascript.info/object-methods](https://javascript.info/object-methods)

[Not working repro](https://www.typescriptlang.org/play?alwaysStrict=false&target=99&jsx=0#code/GYVwdgxgLglg9mABAcwKZQGJzgCgJSIDeAsAFCIWIQIDOUiw2iAvEWZR4gPReJQAWqRHTgAnITX5wQAGwAmiAEZDxAQzkBaBDICeDUXAC2fQVVUyZqUQHIaiAA5Waj6DABuqdpwpzVUVQBcRAC+ADRe3uLA6BD8OIIWcEF0ojBgyAQk5N6c4lAgokhgqADuiAAKBoYwNKg4qjQ6kIg44jRwMh6hiOIAVqjQBMwAfGzZOTkCNQB0JalQqAAifqo4AIwAzBvdawBMG3gA3BETiMFHJxxhl5Q8iCC1Cvaphqqpukp6AkLPqM9wEFQNHaoj4cEQAGtUH9EDIAeZhP4Fog0sImhBECUYAIbhQ7t9EL5-Ch3KgkIo9L84epcZj5ksVjhiiVlv4gmAQIZlKJusyAGpvGCqRSWGjsznczK0jjUMDtSzTOHIJmlVmqXmlAWpYWivC067jM5kE55ApIRhwMjBY2kWV0Ig9VDRKCxM4sFDoLC4C6kMgNdEMcCuBCO52xACSUHwWQ4qhKqmxoZicWsJTE8msetI1t9pCiycj+EOQA)

plunkettscott:
Awesome!

packages/houdini/runtime/network.ts
```diff
@@ -65,14 +70,82 @@ type GraphQLError = {
...
 	params: FetchParams,
 	session?: FetchSession
 ) => Promise<RequestPayload>
 
+// This function is responsible for simulating the fetch context, getting the current session and executing the fetchQuery.
+// It is mainly used for mutations, refetch and possible other client side operations in the future.
+export function executeQuery(
+	artifact: QueryArtifact | MutationArtifact,
+	variables: { [key: string]: unknown },
+	sessionStore: Readable<any>
```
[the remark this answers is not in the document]
pixelmund: `getSession` needs to run on component initialization, because it is using `getContext` under the hood. If we where to call getSession while executing the mutation or query afterwards it would throw an error.

example/src/routes/[filter].svelte
```diff
@@ -83,9 +85,16 @@
 		// trigger the mutation
 		await addItem({ input: { text: inputValue } })
 
+		// try {
```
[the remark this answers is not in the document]
Comment (pixelmund) [reply example/src/routes/[filter].svelte:88]: Yes it was just for testing, i will clean it up

packages/houdini/runtime/network.ts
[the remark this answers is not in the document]
AlecAivazis: Makes sense!

AlecAivazis:
Thanks for getting those changes in so quickly. I'm going to take another pass at this tomorrow but I think it's ready to go.  If you don't mind taking a first pass at some documentation I can review that too otherwise I'll add something once this goes in

packages/houdini/runtime/query.ts
```diff
@@ -86,9 +86,18 @@ export default function query<_Query extends Operation<any, any>>(
...
+				// If new variables are set spread the new variables over the previous ones.
+				if (newVariables) {
+					variableBag = { ...variableBag, ...newVariables }
+				}
+				// Execute the query
+				const result = await executeQuery(artifact, variableBag, sessionStore)
+				// Write the data to the cache
+				writeData(result, variableBag)
+				// Return the data
+				return result.data
```
AlecAivazis, line 100: This shouldn't return anything, right? per the conversation in #57

AlecAivazis (changes requested):

pixelmund:
No problem, always happy to help. I can also review pull requests and help out if you like to add me as a maintainer or something!

packages/houdini/runtime/query.ts
pixelmund: Ah yes, wasn't totally sure about this one.

pixelmund:
I won't have time this week to add documentation, it's the last day before holiday for me.

AlecAivazis:
Sorry this hasn't gotten merged yet! Its been a very busy week for me. I'm going to take my final pass today and get this in ASAP

PR #89, opened 2021-06-02
Document the --pull-schema option

SorenHolstHansen:


README.md
```diff
@@ -159,6 +159,11 @@ npx houdini generate
 
 The generated runtime can be accessed by importing `$houdini` anywhere in your application.
 
+If you have updated your schema on the server, you can pull down the most recent schema by using `--pull-schema` or `-p` like this
```
AlecAivazis, line 162: We probably want to mention that the command will still generate run the runtime. Maybe something like:

> If you have updated your schema on the server, you can pull down the most recent schema before generating your runtime by using `--pull-schema` or `-p`: 
>

AlecAivazis (changes requested):

README.md
SorenHolstHansen: I have updated the description now

AlecAivazis:
Thanks for updating this so quickly!

PR #91, opened 2021-06-08
Fix a bug in pull schema and write the schema to .gql or .graphql files if set in config.

pixelmund:
I noticed that with the current implementation of pull-schema, you have to execute the command twice because the config reads the schema before it gets updated.
I've also updated the writeSchema function to write .gql/.graphql files correctly since the current behaviour writes json into it.


AlecAivazis:
Nice catch!

PR #92, opened 2021-06-09
Add schemaPath option to init command

pixelmund:
This adds a new option to the init command to specify where the schema file should be written to.

### DIFF
--- a/packages/houdini/cmd/init.ts
+++ b/packages/houdini/cmd/init.ts
@@ -8,7 +8,7 @@ import { writeSchema } from './utils/writeSchema'
 export default async (_path: string | undefined) => {
 	// we need to collect some information from the user before we
 	// can continue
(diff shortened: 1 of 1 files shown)


packages/houdini/cmd/init.ts
```diff
@@ -20,6 +20,26 @@ export default async (_path: string | undefined) => {
 			message: 'Are you using Sapper or SvelteKit?',
 			choices: ['Sapper', 'SvelteKit'],
 		},
+		{
+			name: 'schemaPath',
+			type: 'input',
+			default: './schema.json',
```
AlecAivazis, line 26: Let's have the default use the `.graphql` tag since I think its friendlier

AlecAivazis (changes requested):

AlecAivazis:
Thanks for updating that! This should be released now under `v0.6.2`

PR #95, opened 2021-06-13
Query outside routes

### LINKED ISSUE #67: Use without Sapper or SvelteKit
From reading the documentation it's not clear if/ how this can be used within a bare Svelte app.

If this is possible, how would one use Houdini without using Sapper (i.e. only having a simple Svelte app). If this is currently not possible, why not (and how could that be changed, e.g. if I wanted to submit a PR)?

### LINKED ISSUE #72: Tips for debugging failed fetches?
I have not yet had success with fetch queries to my remote graphql server (Hasura). I'm getting a `Cannot read property 'data' of undefined` error, and after adding logging to the environment.js my best guess is that it's not even attempting to fetch since the logging isn't appearing. I'm not clear on the best way to debug this issue, and am hoping a solution here will help others.

My query is adapted straight from the documentation:
```
	import { query, graphql, AllUsers } from '$houdini';
	const { data } = query<AllUsers>(graphql`
		query AllUsers {
			kl_users {
				email
				uid
			}
		}
	`);
```

and the error is:
```
Cannot read property 'data' of undefined
TypeError: Cannot read property 'data' of undefined
    at Module.query (/$houdini/runtime/query.js:15:48)
    at Login.svelte:33:29
    at Object.$$render (<PATH>/node_modules/svelte/internal/index.js:1380:22)
    at eval (/src/lib/layout/Header.svelte:15:115)
    at Object.$$render (<PATH>/node_modules/svelte/internal/index.js:1380:22)
    at eval (<PATH>/src/routes/$layout.svelte:42:94)
    at Object.$$render (<PATH>/node_modules/svelte/internal/index.js:1380:22)
    at root.svelte:31:44
    at $$render (<PATH>/node_modules/svelte/internal/index.js:1380:22)
    at Object.render (<PATH>/node_modules/svelte/internal/index.js:1388:26)
```

### LINKED ISSUE #85: Support for SPA Mode
Would supporting SvelteKit's [`adapter-static` SPA mode](https://github.com/sveltejs/kit/tree/master/packages/adapter-static#spa-mode) be something you are willing to do? Currently the runtime does not work, expecting things like `document.initialValue` to be set but is not because there is no SSR hydration.

[Quite a few people](https://github.com/sveltejs/kit/issues/754), [myself included](https://github.com/sveltejs/kit/issues/754#issuecomment-819011031), embed the static assets into binaries we ship as part of the release process because we use languages that do not support SSR. In my case, that's embedding the assets in a Go application and serving them with an embedded file server.

Original SPA mode PR: https://github.com/sveltejs/kit/pull/1181

AlecAivazis:
This PR adds support for `query` in non-route components as well for applications that aren't using sapper or sveltekit. Along the way, I also added some configuration to the init command that decouples the module type from the framework being used in order to allow for users to opt out of esm if they're using kit, or opt-into them if they are using sapper.

This PR should also add support for kit's SPA mode which builds a bare svelte app out of the project.

closes #67, closes #85, closes #72


pixelmund:
Nice! I'm on vacation for the next 2 weeks so I won't have time to test it but it looks good to me so far.

PR #97, opened 2021-06-18
Write only the changed files

georgecrawford:


packages/houdini/cmd/utils/writeFile.ts
```diff
@@ -0,0 +1,33 @@
...
+function hashString(data: string) {
+	const hashSum = crypto.createHash('sha256')
+	hashSum.update(data)
+	return hashSum.digest('hex')
+}
+
+export async function writeFile(path: string, data: string) {
+	try {
+		const fileHash = await hashFile(path)
+		const dataHash = hashString(data)
```
AlecAivazis, line 20: I might be missing something but do we need to hash the file in order to compare it? Can we just compare the file contents directly?

AlecAivazis (commented):
Thanks for submitting this @georgecrawford! Quick question and we should be able to get this merged in very quickly

packages/houdini/cmd/utils/writeFile.ts
```diff
@@ -0,0 +1,33 @@
...
+
+		if (fileHash === dataHash) {
+			// console.log(`Path ${path} has not changed!`)
+			return
+		}
+	} catch (error) {
+		// ignore
+	}
+
+	console.log(`Path ${path} has changed!`)
```
AlecAivazis, line 30: This will create a very noisy experience when generating the runtime. I'm not sure the user cares to see every file that's changed. I think a better experience would be to only print if an artifact changes (that's what relay does, anyway).

packages/houdini/cmd/utils/writeFile.ts
georgecrawford: Yeah, I see now that I was over-thinking. I've simplified it greatly!

packages/houdini/cmd/utils/writeFile.ts
georgecrawford: Thanks - it was for debugging really, so now removed.

georgecrawford:
@AlecAivazis thanks for the review - I've updated with those tweaks

AlecAivazis (approved):
Looks good! Will merge it in once we get the 👍 from github

AlecAivazis:
Just published this under `v0.7.1`  - thanks for the contribution!

PR #101, opened 2021-06-21
Use Svelte parser

georgecrawford:


AlecAivazis:
I got around the problem you were experiencing but the call to `parse(content)` doesn't handle typescript files and throws an error. i pushed up the work to the `georgecrawford-parse` branch so you can take a look.

After wrestling with this for awhile i think the only real step forward is to rely on the svelte compiler but process the typescript before sending it to the compiler. I tried to use the typescript preprocessor from `svelte-preprocess` for this but ran into some hangups. Will try again tomorrow with a fresh set of eyes if you haven't gotten to it by then.

AlecAivazis:
So after pushing the content through the typescript compiler, the `start` and `end` fields returned from `svelte.parse` are relative to the transpiled version, not the original version which means it can't be used to replace the tag's contents

AlecAivazis:
I spent many hours today trying to get this branch to work, the `parse.test.ts` file passes but once I start to build and run the rest, it all comes crashing down. I'm going to close this in favor of #102 for now and hopefully we can give it another shot later.

PR #117, opened 2021-07-02
Custom scalars

### LINKED ISSUE #56: Custom scalars
Houdini needs a way to specify how to marshal/unmarshal scalars found in documents

AlecAivazis:
This PR adds support for custom scalars, defined by adding a `scalars` key to the `houdini.config.js`. Since this file is imported by the runtime and server code, the imports and logic must be able to run in both environments. This means that users that are relying on `process.env` and other server-only logic will need to use a replace plugin to make it work. This also means that the config file can no longer import `path`. I tried to find an alternative but this was the best compromise I could come up with. 

An example scalar definition: 

```javascript
// houdini.config.js

export default {
	// ...

	scalars: {
		// the name of the scalar we are configuring
		DateTime: {
			// the corresponding typescript type 
			type: 'Date',
			// turn the api's response into that type
			unmarshal(val) {
				return new Date(val)
			},
			// turn the value into something the API can use
			marshal(date) {
				return date.getTime()
			},
		},
	},
}
```

closes #56

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ for the generation of an incredibly lean GraphQL abstraction for your applicatio
     1. [Sapper](#sapper)
     1. [SvelteKit](#sveltekit)
     1. [Svelte](#svelte)
+1. [Config File](#config-file)
 1. [Running the Compiler](#running-the-compiler)
 1. [Fetching Data](#fetching-data)
     1. [Query variables and page data](#query-variables-and-page-data)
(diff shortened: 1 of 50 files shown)


AlecAivazis:
Okay, i think this is ready to go. @pixelmund @thormuller, mind pulling this into your application and giving it a whirl? It should be available as `@next` or version `0.8.6-rc.0`

AlecAivazis:
pixelmund confirmed that his app still runs so i'm going to merge this

Issue #93, opened 2021-06-10
Persisted Queries


pixelmund:
I would like to discuss if we want to support persisted queries in the future. I like how relay handles it and personally i find them quite powerful in terms of security and saving bytes on network requests:

[https://relay.dev/docs/guides/persisted-queries/](https://relay.dev/docs/guides/persisted-queries/)

AlecAivazis:
Its definitely something that I would love to see in `houdini`, just forgot to create the issue 😅

pixelmund:
Thinking about this some more, we could probably use a relay like approach where we create a output file e.g. `queryMap.json` customizable with an cli flag `--persist-output ./src/some_folder/queryMap.json` with the artifact hash as `key` and the query as `value`. We could then pass the artifact or some form to the environments fetchFunction instead of only `text` and `variables`. From there some could simply pass `documentId` (the artifact hash) instead of the full query.

```ts
export default new Environment(async function (operation) {
	// send the request to the api
	const result = await this.fetch('/graphql', {
		method: 'POST',
		headers: {
			'Content-Type': 'application/json',
		},
		body: JSON.stringify({
			// query: operation.text,
			variables: operation.variables,
                        documentId: operation.hash		
            }),
	})

	// parse the result as json
	return await result.json()
})

```

AlecAivazis:
Awesome - in general the approach seems like the right one. I think it's probably more likely that the user is going to follow the standard that urql and apollo both follow which can be found [here](https://github.com/apollographql/apollo-link-persisted-queries#apollo-engine) but I don't think that really matters from our perspective. As long as a hash gets generated and attached to the artifact in a way that the environment has access to when making queries, the user can make that decision for themselves.

Also, `hashDocument` isn't used any more but I left it in the codebase in case we needed it. It's a remnant from when the preprocessor import the artifacts but that's no longer the case.

PR #136, opened 2021-07-16
Persisted Queries #93

pixelmund:
## This implements persisted queries!

I added a `--persist-output` flag to the `generate` command, which can be used to write a `queryMap` to the provided location in json format.

I'm using the `hashDocument` function again and updated it to use `sha256` instead of `md5` since this is what apollo server expects.

I'm passing the hash to the `environments` fetch function, this opens up for 2 different persisted query approaches.

- [Apollo Style](https://github.com/apollographql/apollo-link-persisted-queries#apollo-engine)
- [Relay Style](https://relay.dev/docs/guides/persisted-queries/)

I personally prefer the `--persist-output` option because i'm not using apollo in any ways but this let's the user decide.

The `--persist-output` flag is not required for the `apollo style`.

I'm open to any changes, renames etc.


pixelmund:
Uff the tests are broken, probably because adding the `hash` back to the artifacts. I don't really know how to fix those.

AlecAivazis (changes requested):
Awesome!! Thanks so much for putting this together. It's a bit hard to review with all of the whitespace changes - mind seeing if you can fix them? It looks like jest is leaving down spaces instead of tabs

pixelmund:
Yeah sure i will fix them on sunday when i got some free time. I had issues with jest and tabs, for some reason switching to spaces on these Files fixed it. But im sure it will work with Tabs too.

pixelmund:
I will keep this pr as draft until pagination is merged, because there are many changes to the tests.

AlecAivazis:
Sounds good! the pagination branch also dramatically simplifies the artifact generation logic. Instead of constantly wrestling ASTs it builds up javascript objects and then coverts it to the proper AST in a single step. That lets us share types between the runtime and generation code to make sure everything matches up.

Issue #137, opened 2021-07-17
Node version


SorenHolstHansen:
I see that houdini requires node version 15 or higher to work, however hosting providers like vercel only go as high as node version 14.x. Is there a reason for needing 15 or higher, otherwise it would be great to allow for lower versions

AlecAivazis:
This `engines` field was set in order to avoid an `Unexpected token *` exception that was reported in #103. It's caused by an `import * from ...` statement in the built version of the cli. To be honest, I didn't really investigate if there were versions less than `15.0.0` that supported this syntax but it's definitely important to be able to support vercel. If you have the time to find a version that works, I'd really appreciate it. Otherwise I can investigate it a bit more once I'm done with pagination.

SorenHolstHansen:
I can maybe make the time tomorrow, will see

SorenHolstHansen:
How do I reproduce the issue in #103, am unsure if the issue is not reintroduced when I change stuff.

AlecAivazis:
I think the only way to reproduce it is to downgrade your node version (with something like `nvm`) to find the lowest version that houdini supports. If I remember correctly, the error would happen when invoking the `init` function but it's probably safest to also make sure that the `generate` command also works in a given version. You can use the example app if you need something to test against. If the minimal version isn't supported by vercel, we'll have to figure out how to tweak the build configuration to compile the sourcecode into something that can run on their environment.

AlecAivazis:
Closing this since #139 has been merged. Thanks again @SorenHolstHansen

Issue #103, opened 2021-06-24
Unexpected token *


TheBeachMaster:
Getting started with Houdini on Sapper throws the following error 

`Unexpected token *`

## Steps:
In an existing sapper project
1. `yarn add -D houdini houdini-preprocess`
2. `npx houdini init`

## Result:

`Unexpected token *`

System Info

```
  System:
    OS: macOS 11.4
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 1.54 GB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  npmPackages:
    @babel/core: ^7.13.16 => 7.13.16 
    @babel/plugin-syntax-dynamic-import: ^7.8.3 => 7.8.3 
    @babel/plugin-transform-runtime: ^7.13.15 => 7.13.15 
    @babel/preset-env: ^7.13.15 => 7.13.15 
    @babel/runtime: ^7.13.17 => 7.13.17 
    @rollup/plugin-babel: ^5.3.0 => 5.3.0 
    @rollup/plugin-commonjs: ^18.0.0 => 18.0.0 
    @rollup/plugin-node-resolve: ^11.2.1 => 11.2.1 
    @rollup/plugin-replace: ^2.4.2 => 2.4.2 
    @rollup/plugin-url: ^6.0.0 => 6.0.0 
    compression: ^1.7.4 => 1.7.4 
    houdini: ^0.7.1 => 0.7.1 
    houdini-preprocess: ^0.7.1 => 0.7.1 
    npm-check-updates: ^11.5.1 => 11.5.1 
    polka: next => 1.0.0-next.14 
    qrcode: ^1.4.4 => 1.4.4 
    qrcode-svg: ^1.1.0 => 1.1.0 
    rollup: ^2.45.2 => 2.45.2 
    rollup-plugin-svelte: ^7.1.0 => 7.1.0 
    rollup-plugin-terser: ^7.0.2 => 7.0.2 
    sapper: ^0.29.1 => 0.29.1 
    sirv: ^1.0.11 => 1.0.11 
    svelte: ^3.37.0 => 3.37.0 
    svelte-qrcode: ^1.0.0 => 1.0.0 
```

AlecAivazis:
Hey @TheBeachMaster - sorry you ran into this.

Is there a stack trace associated with the error?

TheBeachMaster:
Hello @AlecAivazis  unfortunately there's none

[...]

PR #139, opened 2021-07-20
Use node version >=14.13

SorenHolstHansen:
It seems that node version >=14.13 does not reintroduce the problems in #103, and allows for support for Vercel as discussed in #137.
Have tried version 14.12 and below, and there are issues with the graphql Kind export, and maybe others, as can be seen in the CI checks for the first commit.


AlecAivazis:
Perfect! Thanks for hunting this down

PR #144, opened 2021-07-24
Pagination

### LINKED ISSUE #55: Pagination
I'm not quite sure what the API should look like, but there should be some kind of support for interacting with connections that have the usual pagination args.

### LINKED ISSUE #123: @when should look like @with
The `@when` directive's api is a bit clunky and can only provide one value at a time. It should be updated to look like `@with` or `@arguments` now that we have internal utilities for  understanding the directives without strict schemas

AlecAivazis:
This PR adds pagination support to houdini 🎉 

At a high level, pagination is driven by two new functions (`paginatedQuery` and `paginatedFragment`) and a new `@paginate` directive. Both functions return an object containing `loadNextPage`, `loadPreviousPage`, and `pageInfo` along with the usual `data` and `loading` stores . The fragment support for pagination relies on the same `Node` query and interface that relay does if it's not a fragment on the schema's query type. `paginatedQuery` works without any extra constraints on the api. An example:

```svelte
<script>
  const {  data, loading, loadNextPage, pageInfo } = paginatedQuery(graphql`
      query UserList { 
          friends(first: 10) @paginate {
              edges { 
                   node { 
                       id
                   }
              }
          }
      }
  `)
</script>

{#if $pageInfo.hasNextPage}
    <button onClick={() => loadNextPage()}> load more </button>
{/if}
```

For more information, you can check out the README. 

Along the way, I also improved the artifact generation logic to avoid relying so heavily on recast's AST builder. Instead it builds up a javascript object (using the typedefinitions from the runtime) and then converts it to the appropriate AST before writing the file. In order to make this simplier, `when` and `when_not` were also updated to an api structure similar to the `@with` and `@arguments`.  

This branch is published under `houdini@pagination` so people can start testing it before I merge..

closes #55, closes #123

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,10 +23,10 @@ jobs:
               run: |
                   git checkout ${{ github.event.pull_request.base.ref }}
 
-            - name: Setup Node 15
+            - name: Setup Node 14
               uses: actions/setup-node@v1
               with:
-                  node-version: 15
+                  node-version: 14.13
                   registry-url: 'https://npm.pkg.github.com'
 
             - name: Set Git Identity
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,10 +18,10 @@ jobs:
         runs-on: ubuntu-latest
 
         steps:
-            - name: Use Node.js 15.x
+            - name: Use Node.js 14.13
               uses: actions/setup-node@v1
               with:
-                  node-version: 15.x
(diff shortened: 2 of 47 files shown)


AlecAivazis:
Thanks @pixelmund for testing this for me! @thormuller do you think you could install `houdini@pagination` and make sure everything still works for you?

AlecAivazis:
Just received confirmation from thor that his app is still working as expected - going to merge this now 🎉

PR #145, opened 2021-07-25
Feat: Export type ConfigFile to enable typable configs + code completion

benbender:
Hey,

just a small enhancement to enable typable configs. It's quite nice to have some confidence while changing configs and comes with basically zero cost...

Btw, thanks for your work - its great to expand the svelte-way of tackle modern frontend-problems to the graphql-client! :)

### DIFF
--- a/example/houdini.config.js
+++ b/example/houdini.config.js
@@ -1,4 +1,5 @@
-export default {
+/** @type {import('houdini/build/runtime').ConfigFile} */
+const config = {
 	schemaPath: './schema/schema.gql',
 	sourceGlob: 'src/**/*.svelte',
 	framework: 'kit',
@@ -16,3 +17,5 @@ export default {
 		},
 	},
 }
+
+export default config
--- a/example/houdini.config.js
+++ b/example/houdini.config.js
@@ -1,4 +1,4 @@
-/** @type {import('houdini/build/runtime').ConfigFile} */
+/** @type {import('houdini').ConfigFile} */
 const config = {
 	schemaPath: './schema/schema.gql',
 	sourceGlob: 'src/**/*.svelte',
--- a/example/houdini.config.js
+++ b/example/houdini.config.js
@@ -1,5 +1,5 @@
 /** @type {import('houdini').ConfigFile} */
-const config = {
+export default {
 	schemaPath: './schema/schema.gql',
 	sourceGlob: 'src/**/*.svelte',
 	framework: 'kit',
@@ -17,5 +17,3 @@ const config = {
 		},
 	},
 }
-
-export default config
--- a/packages/houdini/cmd/init.ts
+++ b/packages/houdini/cmd/init.ts
@@ -141,12 +141,11 @@ const configFile = ({
 
 	return module === 'esm'
 		? // SvelteKit default config
-		  `
-		  /** @type {import('houdini').ConfigFile} */
-		  export default ${configObj}
+		  `/** @type {import('houdini').ConfigFile} */
+export default ${configObj}
 `
 		: // sapper default config
 		  `/** @type {import('houdini').ConfigFile} */
-		  module.exports = ${configObj}
+module.exports = ${configObj}
 `
 }
--- a/packages/houdini/cmd/init.ts
+++ b/packages/houdini/cmd/init.ts
@@ -141,9 +141,16 @@ const configFile = ({
 
 	return module === 'esm'
 		? // SvelteKit default config
-		  `export default ${configObj}
+		  `
+		  /** @type {import('houdini').ConfigFile} */
+		  const config = ${configObj}
+
+		  export default config;
 `
 		: // sapper default config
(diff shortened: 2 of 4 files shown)


AlecAivazis:
Oh nice! I like this a lot. I think we can clean it up a bit by exporting the type from `houdini/packages/houdini/cmd/types.ts` instead of the runtime. That should let users type their config file with:

```javascript
const config = {
   ...
}
```

benbender:
Even cleaner and nicer :) Do you want to do it or should I update the PR?

AlecAivazis:
You can do it if you want otherwise I can pick up the torch between work this week

AlecAivazis:
What do you think about adding this comment to the default config file stamped out by `init`? It seems like a useful thing to add in general and if i remember correctly, sveltekit also includes the type comment by default. Fwiw, that's defined [here](https://github.com/AlecAivazis/houdini/blob/main/packages/houdini/cmd/init.ts#L142-L148)

benbender:
I think that everything should be typed in general - and that should more people should be educated that JS also can be typed (with a reasonable editor). So I don't see any (potential) downside in this change and would vote in favor!

AlecAivazis:
Sweet - feel free to update your PR and then we can merge this in 👍

benbender:
@AlecAivazis Just had a look and I'm unshure how to tackle this. If I understood the code correctly, the config-file on `houdini init` is generated here https://github.com/AlecAivazis/houdini/blob/c990e0ae66e345e54335d2c6a16ad9e2f322cb9f/packages/houdini/cmd/init.ts#L85 from an object. I don't have a clou how to add a comment without rewriting the whole infrastructure to be template-based...

AlecAivazis:
@benbender - like I mentioned above, the actual contents for the config file are [here](https://github.com/AlecAivazis/houdini/blob/main/packages/houdini/cmd/init.ts#L142-L148) so you just have to add the comment to both strings so that it shows up in the esm and cjs config files

benbender:
Nevermind, just saw it - I overlooked the function below and assumed there was some kind of external package generating the file. PR updated...

packages/houdini/cmd/init.ts
```diff
@@ -141,9 +141,16 @@ const configFile = ({
 
 	return module === 'esm'
 		? // SvelteKit default config
-		  `export default ${configObj}
+		  `
+		  /** @type {import('houdini').ConfigFile} */
+		  const config = ${configObj}
+
+		  export default config;
```
AlecAivazis, line 148: Do we need to define the variable in a separate line from the export? The [sveltekit default config](https://github.com/AlecAivazis/houdini/blob/main/example/svelte.config.js) doesn't seem to do that. I'd prefer to keep it on one line if possible

AlecAivazis (changes requested):

packages/houdini/cmd/init.ts
benbender: I like it that way so you get the documentation/preview/etc (don't know the exact name of those features) by your editor when you hover the var. That's the reason why I normally assign configs first before exporting them...

AlecAivazis: Let's still have it on one line since it seems like an IDE-specific optimization for a particular workflow. For example, I get a very similar experience with ctrl+spacebar in vscode

benbender: Updated

example/houdini.config.js
```diff
@@ -1,4 +1,5 @@
-export default {
+/** @type {import('houdini').ConfigFile} */
+const config = {
```
AlecAivazis, line 2: Mind updating the example project aswell? I think it's important to keep that in sync

AlecAivazis (changes requested):

example/houdini.config.js
benbender: done

packages/houdini/cmd/init.ts
```diff
@@ -141,9 +141,12 @@ const configFile = ({
 
 	return module === 'esm'
 		? // SvelteKit default config
-		  `export default ${configObj}
+		  `
```
AlecAivazis, line 144: Sorry for submitting these requests one at a time. Hard to juggle this between work 😅 

I'm not sure this will render correctly - i think you want the comment at the start of the string and then the export at the beginning of the line without any of the white space. You'll also need to update the tests

AlecAivazis (changes requested):

packages/houdini/cmd/init.ts
benbender: Wouldn't do any harm afaik, but fixed anyways.

AlecAivazis:
Just tested this locally before publishing - you were right, in order to get intellisense we need to add the comment on the variable definition. It doesn't work to comment on the export statement. Going to update this to reflect how you originally had it

PR #149, opened 2021-07-29
Feat: onLoad Hook

### LINKED ISSUE #147: Add hook for executing logic at the beginning of load
A user might want to perform additional logic in the generated load function. One example brought up by @pixelmund was to ensure that the user was authenticated a perform and perform a redirect if not. We might want to explore adding a life cycle method that we can invoke in the generated `load`. 

Another option might be to bring back some old code that tried to inject the `load` logic into an existing function. This approach is dramatically more complicated to implement but also allows the user to add additional loading logic that might not fit within their graphql api

pixelmund:
Closes #147

This implements the `onLoad` hook, it get's the same args as the variable functions and can return `this.error` `this.redirect` etc.

Also if you return some object from the function it get's added to the return of `load`.

e.g.

```svelte
  <script context="module">
     // It has access to the same arguments and this.error this.redirect as the variable functions
      export function onLoad({page, session}){

       if(!session.authenticated){
            return this.redirect(302, '/login')
       }

         return {  hello: "world" }
      } 
 </script>

 <script>
     export let hello
 </script>
```

### DIFF
--- a/packages/houdini-preprocess/src/transforms/query.test.ts
+++ b/packages/houdini-preprocess/src/transforms/query.test.ts
@@ -507,10 +507,16 @@ test('onLoad hook', async function () {
 			export async function onLoad(){
 			   return this.redirect(302, "/test")
 			}
+
+			export function TestQueryVariables(page) {
+				return {
+					test: true
+				}
+			}
 		</script>
 		<script>
 			const { data } = query(graphql\`
-				query TestQuery {
+				query TestQuery($test: Boolean!) {
 					viewer {
 						id
 					}
@@ -530,13 +536,29 @@ test('onLoad hook', async function () {
 		    return this.redirect(302, "/test");
 		}
 
+		export function TestQueryVariables(page) {
+		    return {
+		        test: true
+		    };
+		}
+
 		export async function load(context) {
 		    const _houdini_context = new RequestContext(context);
-		    const _TestQuery_Input = {};
 
 		    await _houdini_context.onLoadHook({
 		        "mode": "sapper",
 		        "onLoadFunction": onLoad
+		    });
+
+		    if (!_houdini_context.continue) {
+		        return _houdini_context.returnValue;
+		    }
+
+		    const _TestQuery_Input = _houdini_context.computeInput({
+		        "config": houdiniConfig,
+		        "mode": "sapper",
+		        "variableFunction": TestQueryVariables,
+		        "artifact": _TestQueryArtifact
 		    });
 
 		    if (!_houdini_context.continue) {
--- a/packages/houdini-preprocess/src/transforms/query.test.ts
+++ b/packages/houdini-preprocess/src/transforms/query.test.ts
@@ -500,75 +500,70 @@ describe('query preprocessor', function () {
 	test.todo('adds arguments to an empty preload')
 })
 
-// test('onLoad hook', async function () {
-// 	const doc = await preprocessorTest(
-// 		`
-// 		<script context="module">
-// 			export async function onLoad(){
-// 			   return this.redirect(302, '/test')
-// 			}
-// 		</script>
-// 		<script>
-// 			const { data } = query(graphql\`
-// 				query TestQuery {
-// 					viewer {
-// 						id
-// 					}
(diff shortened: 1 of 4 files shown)


packages/houdini-preprocess/src/transforms/query.test.ts
```diff
@@ -499,3 +499,71 @@ describe('query preprocessor', function () {
...
+
+		export async function onLoad() {
+		    return this.redirect(302, "/test");
+		}
+
+		export async function load(context) {
+		    const _houdini_context = new RequestContext(context);
+		    const _TestQuery_Input = {};
+
+		    _houdini_context.onLoadHook({
```
AlecAivazis, line 537: Since `onLoad` is async, do we need to `await` this here?

packages/houdini-preprocess/src/transforms/query.ts
```diff
@@ -472,16 +497,32 @@ function addKitLoad(config: Config, body: Statement[], queries: EmbeddedGraphqlD
...
-			AST.objectProperty(AST.identifier(preloadKey), AST.identifier(preloadKey)),
-			// @ts-ignore
-			AST.objectProperty(
-				AST.identifier(variableIdentifier),
-				AST.identifier(variableIdentifier)
+		if (onloadDefinition) {
+			// add the field to the return value of preload
+			propsValue.properties.push(
+				// @ts-ignore
+				AST.spreadProperty(
```
AlecAivazis, line 504: I think it would be cleaner if you check for the `onLoadDefinition` before pushing the other values so we don't have to duplicate the actual query responses. Something like

```javscript
if (onLoadDefinition) {
    propsValue.properties.push(...)
}

propsValue.properties.push( ...)
```

AlecAivazis (changes requested):

packages/houdini-preprocess/src/transforms/query.ts
pixelmund: You're right😅

packages/houdini-preprocess/src/transforms/query.test.ts
pixelmund: Yeah probably

AlecAivazis:
We also probably want to add some documentation for this hook. I can take a stab at it after this goes in if you prefer

packages/houdini/runtime/network.ts
```diff
@@ -233,6 +233,33 @@ export class RequestContext {
...
+			mode === 'kit'
+				? await (onLoadFunction as KitLoad).call(this, this.context)
+				: await (onLoadFunction as SapperLoad).call(
+						this,
+						this.context.page,
+						this.context.session
+				  )
+
+		// If the result is null or undefined, or the result isn't an object return early
+		if (result == null || typeof result !== 'object') {
```
AlecAivazis, line 256: I think we also want to check for `this.continue` here incase the user returns an error from `onLoad`

packages/houdini-preprocess/src/transforms/query.test.ts
```diff
@@ -499,3 +499,71 @@ describe('query preprocessor', function () {
...
+		export async function load(context) {
+		    const _houdini_context = new RequestContext(context);
+		    const _TestQuery_Input = {};
+
+		    await _houdini_context.onLoadHook({
+		        "mode": "sapper",
+		        "onLoadFunction": onLoad
+		    });
+
+		    if (!_houdini_context.continue) {
```
AlecAivazis, line 542: I think the way this is currently implemented might cause confusing behavior if there is a variable function and `onLoad` defined in the same component. We want to check for `context.continue` after each step so that an error in `onLoad` will prevent the variable function from firing

AlecAivazis (changes requested):

packages/houdini-preprocess/src/transforms/query.test.ts
pixelmund: Yeah that makes sense

AlecAivazis (approved):
Sweet! Thanks for getting this out so quickly

PR #153, opened 2021-08-06
fix leaking cache, while ssr

pixelmund:
Currently the cache leaks through requests, which could lead to huge cache size and maybe some data leaks? 

You can reproduce the issue if you import the cache inside `hooks` `handle` and log `cache._data` it' always is filled.

 I think this PR will  also help implementing the fetch-policy since data is always really cached on the client side.

This approach uses sveltes get/setContext which creates a new Cache for every request and keeps only one cache on the client side. 
This would introduce a new setCache() function which should be invoked in the entry point file e.g. `__layout.svelte` or `App.svelte`. 

Maybe there is a better naming for setCache(), it's a bit weird for me needing to invoke `setEnvironment` in `context module` and `setCache` in `instance script` - Also this is a breaking change and needs proper documentation or probably some kind of error message if the context is not found

### DIFF
--- a/example/src/routes/__layout.svelte
+++ b/example/src/routes/__layout.svelte
@@ -1,12 +1,19 @@
 <script context="module">
 	import env from '../environment'
 	import { setEnvironment, setCache, getCache } from '$houdini'
+	import { browser } from '$app/env'
 
 	setEnvironment(env)
 </script>
 
 <script>
 	setCache()
+	const cache = getCache()
+
+	if (browser) {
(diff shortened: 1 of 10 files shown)


AlecAivazis:
Hey @pixelmund - thanks for opening this! I think I’m actually going to close this and bring some of it into my fetch-policy branch. I think for now, having the cache always disabled on the server is enough for fetch-policy which is an client-only feature for now

PR #157, opened 2021-08-19
Fetch policy

AlecAivazis:
This PR gives the user the ability to specify the cache policy for a particular query. It does this with a new `@cache` decorator which looks like:

```graphql
query AllItems @cache(policy: CacheOrNetwork) {
    ...
}
```

There are 3 different policies that can be specified:

- **CacheOrNetwork**:  will first check if a query can be resolved from the cache. If it can, it will return the cached value and only send a network request if it wasn't present.
- **CacheAndNetwork**: will use cached data if it exists and always send a network request after the compmonent has mounted to retrieve the latest data
- **NetworkOnly**: will never check if the data exists in the cache and always send a network request

This PR also implements garbage collection for the cache which is based on a simple hereustic that will retain data for a configurable number of queries (default 10). For a concrete example, consider the example app that has 3 routes. If you load one of the routes and then click between the other two 5 times, the first route's data will still be resolvable (and the counter will reset). If you then toggle between the other routes 10 times and then try to load the first route, a network request will be sent. This number is configurable with the `cacheBufferSize` config value (name pending). 

todo:

-  [x] Document policies
-  [x] CacheAndNetwork in the presence of pagination should retain the full loaded page
-  [x] Make default cache policy customizable 
-  [x] avoid double query with CacheAndNetwork if initial load was loaded from the network


pixelmund:
Awesome! If you like me to test it you can publish under `next` and i will see once i got some free time.

AlecAivazis:
Fantastic! There are still some small things I would like to address before doing that but I should have something in a few days that's ready for you.

AlecAivazis:
@pixelmund i think its ready for you to test. You should be able to get the version with `@next`

pixelmund:
@AlecAivazis Nice! Hope i can get to it soon.

pixelmund:
I've just tested it in my project and as far as i can tell it is working fine! I've tried NetworkOnly and CacheOrNetwork for now.

AlecAivazis:
That's a little terrifying and exciting at the same time haha. Thanks for checking it out! I'll give it a few more days of poking at my end in case one of us finds something before merging

pixelmund:
You're welcome! Yeah, don't have that much time right now, but i can try to play around a little more soon!

AlecAivazis:
@pixelmund - mind pulling this down and giving it another shot? I finally addressed the last bit I wanted to get in before merging this - calling `refetch` now reloads the entire current page

pixelmund:
@AlecAivazis sure but i don't know when i will get to it.

AlecAivazis:
Okay, i finally got around to fixing an issue I found with `CacheAndNetwork`. I think this is good to go but i'm going to let it simmer for a few more days.

PR #164, opened 2021-08-23
Update yarn dev command in example project to run both web & api servers

ledesmablt:
Was thinking that this would be a nice little quality of life improvement for contributors.

image


AlecAivazis:
Thanks for this!

Issue #162, opened 2021-08-23
houdini generate -p throws error if schema file doesn't exist


ledesmablt:
Use case: I want to run the `generate` command in CI and I don't want to commit the `$houdini` folder and `schema.graphql` file to minimize clutter from generated files in my repo.

I already have an existing `houdini.config.js` file. When I run `houdini generate -p`, I'd expect it to pull the latest schema regardless of whether or not an existing schema file exists. Instead, if the schema file doesn't exist, it gives me this message:

```
(node:25587) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, open
 '/Users/benj/projects/preview-app/schema.graphql'
    at Object.openSync (fs.js:498:3)
    at Object.readFileSync (fs.js:394:35)
    at new Config (file:///Users/benj/projects/preview-app/node_modules/houdini/build/cmd.js:
689:64)
    at file:///Users/benj/projects/preview-app/node_modules/houdini/build/cmd.js:1048:31
    at step (file:///Users/benj/projects/preview-app/node_modules/houdini/build/cmd.js:346:23
)
    at Object.next (file:///Users/benj/projects/preview-app/node_modules/houdini/build/cmd.js
:327:53)
    at fulfilled (file:///Users/benj/projects/preview-app/node_modules/houdini/build/cmd.js:3
17:58)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:25587) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error origin
ated either by throwing inside of an async function without a catch block, or by rejecting a
promise which was not handled with .catch(). To terminate the node process on unhandled promi
se rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cl
i.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:25587) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In th
e future, promise rejections that are not handled will terminate the Node.js process with a n
on-zero exit code.
✨  Done in 0.25s.
```

I suspect that this is caused by the `getConfig()` function depending on the schema file already existing ([see code](https://github.com/AlecAivazis/houdini/blob/main/packages/houdini-common/src/config.ts#L92-L103)). I understand that `houdini init` creates this file, but when running the whole thing in CI I'd want a way to automatically create the schema file since `init` requires user input through the houdini CLI.

[...]

PR #165, opened 2021-08-23
Address houdini generate -p errors if schema file doesn't exist

ledesmablt:
Bug fixes for #162

```
benj@MacBook ~/projects/houdini/example
fix/issue-162* $ ls schema
index.cjs

benj@MacBook ~/projects/houdini/example
fix/issue-162* $ yarn generate
Error: Schema file does not exist! Create it using houdini generate -p
    at new Config (file:///Users/benj/projects/houdini/packages/houdini/build/cmd.js:688:23)
    at file:///Users/benj/projects/houdini/packages/houdini/build/cmd.js:1067:31
    at step (file:///Users/benj/projects/houdini/packages/houdini/build/cmd.js:346:23)
    at Object.next (file:///Users/benj/projects/houdini/packages/houdini/build/cmd.js:327:53)
    at fulfilled (file:///Users/benj/projects/houdini/packages/houdini/build/cmd.js:317:58)

benj@MacBook ~/projects/houdini/example
fix/issue-162* $ yarn generate -p
Pulled latest schema from http://localhost:4000/graphql
ItemEntry_item
UncompleteItem
ItemUpdate
AllItems
AddItem
NewItem
CompleteItem
DeleteItem
```

### DIFF
--- a/packages/houdini/cmd/main.ts
+++ b/packages/houdini/cmd/main.ts
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
 // external imports
-import { getConfig } from 'houdini-common'
+import { getConfig, loadConfigFile } from 'houdini-common'
 import { Command } from 'commander'
 import path from 'path'
 // local imports
@@ -21,14 +21,12 @@ program
 	.action(
 		async (args: { pullSchema: boolean; persistOutput?: string } = { pullSchema: false }) => {
 			// grab the config file
-			const config = await getConfig()
+			let config
 
 			try {
-				if (args.persistOutput) {
-					config.persistedQueryPath = args.persistOutput
-				}
 				// Pull the newest schema if the flag is set
 				if (args.pullSchema) {
+					config = await loadConfigFile()
 					// Check if apiUrl is set in config
 					if (!config.apiUrl) {
 						throw new Error(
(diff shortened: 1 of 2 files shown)


packages/houdini/cmd/main.ts
```diff
@@ -38,14 +36,19 @@ program
...
 					// Write the schema
-					const schema = await writeSchema(
+					await writeSchema(
 						config.apiUrl,
 						config.schemaPath
 							? config.schemaPath
 							: path.resolve(targetPath, 'schema.json')
 					)
-					// Set the newly written schema into the config
-					config.schema = schema
```
AlecAivazis, line 48: I think this line was dropped, i dont see an equivalent in the rest of the diff

AlecAivazis (changes requested):

packages/houdini/cmd/main.ts
ledesmablt: it's on line 49: `config = await getConfig()`

instead of calling `getConfig()` prior to the schema pull, it's called after that. this will load the latest pulled schema that was just written, so no need to overwrite it in the config variable

AlecAivazis: Ahhh, right - thanks!

AlecAivazis (approved):
works like a charm!

PR #172, opened 2021-08-28
Nested error logging on generate command

ledesmablt:
If generate error has nested objects, log their contents

Before
image

After
image


AlecAivazis:
Awesome! Thanks!

Issue #170, opened 2021-08-25
[Preprocessor] detecting isRoute broken on windows


pixelmund:
Hey i noticed that if i run a project on windows, the preprocessor is not correctly detecting `isRoute`.

```js
const isRoute =
  config.framework !== "svelte" &&
  !config.static &&
  doc.filename.startsWith(path.join(config.projectRoot, "src", "routes"));

```

The startsWith on windows always fails because there is a mismatch of `/` and `\`

e.g. `C:/Users/user/DEV/src/routes/index.svelte C:\Users\user\DEV\src\routes`

If i change the code to this it works:
```js
const isRoute =
  config.framework !== "svelte" &&
  !config.static &&
  path
    .resolve(doc.filename)
    .startsWith(path.join(config.projectRoot, "src", "routes"));
```

I don't have the time to open a PR currently.

AlecAivazis:
Awesome! Thanks for providing the fix too 👍 I'll sneak it into #171

pixelmund:
Hey i think this is still broken.. The fix never made it into that PR as i can tell. Was playing around with a new project and the preprocessor doesn't know if its a route file or component.

PR #223, opened 2022-01-30
Fix isRoute detection in preprocessor

pixelmund:
I've stumbled across the #170 issue again on windows, i've gone ahead and found a simple fix, windows paths always contain back slashes, but the doc.filename forward slashes so to make the detection correct we replace all backlashes with forward slashes.


AlecAivazis:
Thanks for fixing this @pixelmund! Nice to hear you're still using houdin

PR #228, opened 2022-02-22
Cache v2

AlecAivazis:
This PR reorganizes the cache into more discrete parts in order to improve the maintainability as well as provide a way to swap out different parts of the system to allow for improved experimentation. Along the way, I built the infrastructure for supporting optimistic responses for mutations but I want some more time to think through that API so this PR does not include a way for someone to specify them. I wanted to get this out ASAP so people could test it and find anything that pops up. I am going to test some apps that I have with these changes and then deploy it so people can give it a shot once I've worked through the initial wave of findings.

All of the old tests have been kept and still pass with the new implementation which gives me a wary sense of comfort. I tried to reduce the changes to the actual public API as best I could but if someone was using anything through the old `internal` attribute, they will likely run into issues. In order to make the contract/abstraction boundary more clear, I moved all of the implementation details/internal methods into another class so that someone who is exploring the cache's public API with `console.log` only sees 7 things: read, write, subscribe, unsubscribe, list, delete, and `_internal_unstable` which they hopefully will understand shouldn't be relied on


AlecAivazis:
This version has now been published under the `@next` tag if anyone has the time to check it out. 

cc @pixelmund @thormuller @SorenHolstHansen @plunkettscott

PR #229, opened 2022-02-25
Make result type possibly undefined

SorenHolstHansen:
I am always forgetting that `$data` might be `undefined` since the typescript type does not specify it, so i changed the result type to be a union of the old type and `undefined`. I was not too sure if i should put it on the Query type, `$result` type or on the runtime query, mutation, etc. data types

### DIFF
--- a/packages/houdini/cmd/generators/typescript/index.ts
+++ b/packages/houdini/cmd/generators/typescript/index.ts
@@ -151,7 +151,12 @@ async function generateOperationTypeDefs(
 					readonlyProperty(
 						AST.tsPropertySignature(
 							AST.stringLiteral('result'),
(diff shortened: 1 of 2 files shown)


AlecAivazis:
Thanks for doing this @SorenHolstHansen!

Your PR has made me consider if `null` is a better default value for a store that hasn't been preloaded. What do you think?

SorenHolstHansen:
@AlecAivazis that is a good question. I am not a fan of js having both null and undefined, and usually just stick whatever i like on the variable, so i am not the best person to answer that question ;)

AlecAivazis:
Okay, no problem. We can leave it as undefined now since that's the current behavior and revisit it later

Issue #232, opened 2022-03-12
Make load hooks more versatile


fehnomenal:
On the page I'm building with kit I need to display categories for a navigation component and want to highlight the active category (the currently viewed category or the category of the currently viewed entry). I see 3 possible approaches:

1. I could just retrieve the categories on each page where I need them, extract somehow the active category and give those as props to the `Navigation` component. This would mean much duplicated code and I think I would need to find an unique name for each query, right? Also it would be not good for caching, right?
2. Retrieve the categories in `__layout.svelte`, put them into a context and extract them where needed.
3. Retrieve the categories in `__layout.svelte` and pass them to further pages via the `stuff` property. This would mean that the `afterLoad` hook may return an `LoadOutput` object.

I think the third option would be the best (at least for kit, I did not look into how the kit code is wrapped for sapper) and implemented a PoC which works alright. Basically my modified code generates this as the `load` return:
```js
  const _houdini_props = {
    _GetCategoriesPublic,
    _GetCategoriesPublic_Input: {},
    _GetCategoriesPublic_Source
  };
  if ("props" in _houdini_context.returnValue)
    return {
      ..._houdini_context.returnValue,
      props: {
        ..._houdini_context.returnValue.props,
        ..._houdini_props
      }
    };
  return {
    props: {
      ..._houdini_context.returnValue,
      ..._houdini_props
    }
  };
```

The fallback means I have to return something for the `props` key in `__layout.svelte`s `afterLoad`:
```js
    return {
      props: {},
      stuff: {
        categories: data.GetCategoriesPublic.categories,
      },
    };
```

Would you be willing to implement something like this/accept a PR for this?

Then I would polish the code a bit and also handle returns from both `beforeLoad` and `afterLoad`.

AlecAivazis:
Hey @fehnomenal! I'm a little distracted at the moment with some other work for Houdini so I haven't found the time to look at your proposal but I will get to it as soon as I have some time. I didn't want you to think I was ignoring you. Thanks for being patient!

AlecAivazis:
I think you're right that the third option is probably best and could be a useful improvement for many other use cases. Feel free to open a PR if you're still interested, i'd love to take a look :)

fehnomenal:
Glad you like it and that I can contribute to this great project :)

I will prepare a PR but will probably need your help for adding some test cases :sweat_smile:

AlecAivazis:
[...]

PR #238, opened 2022-03-21
Enhance hook function returns

fehnomenal:
See #232

Maybe I went a bit overboard but I restructured the code generation a little bit to make it easier to work for me. I think it is best to review the commits in isolation as otherwise the changes to the generated code in the tests are rather confusing.

1. In commit 35f2be2c04053f344f926f2f670268ed08c41136 I removed some unneeded checks and changed the order of handling query documents, so it will match the order as defined in the svelte code.
2. In commit 867751f7df30e80081488772e6b73db38b7d5592 I took the liberty of (hopefully) improving the performance by not fetching sequentially but in parallel.
3. Commit 9ab933651d44eeab663e280c12c1e23ced60d003 is finally the meat of the PR and assigns hook results to variables, merges them if necessary and also merges `props` and `stuff`. I thought the other keys from `LoadOutput` can already be set by calling `this.redirect()` and `this.error()`.

I use this currently in a sveltekit project and it works as expected, but I cannot say anything about sapper. The generated code is affected but I guess since `convertKitPayload` simply extracts `props` in the happy path everything should be alright.

### DIFF
--- a/packages/houdini-preprocess/src/transforms/query.test.ts
+++ b/packages/houdini-preprocess/src/transforms/query.test.ts
@@ -1236,12 +1236,31 @@ test('both beforeLoad and afterLoad hooks', async function () {
 
 		    const afterHookReturn = _houdini_context.returnValue;
 
+		    const hookReturn = {
+		        ...beforeHookReturn,
+		        ...afterHookReturn
+		    };
+
+		    if (hookReturn.props) {
+		        hookReturn.props = {
+		            ...beforeHookReturn.props,
+		            ...afterHookReturn.props
+		        };
+		    }
+
+		    if (hookReturn.stuff) {
+		        hookReturn.stuff = {
+		            ...beforeHookReturn.stuff,
+		            ...afterHookReturn.stuff
+		        };
+		    }
+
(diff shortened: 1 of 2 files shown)


AlecAivazis:
Hey @fehnomenal, thanks for submitting this. Would you be willing to split up the PR into a few different ones so I can more easily track the changes? I would also like to keep the logic as close to the original version as possible since it has been vetted by many people.  I'm happy to see that multiple requests would happen in parallel but I'd like to review the load changes and the parallelism separately

fehnomenal:
Hey! I did as you requested and got rid of the first two commits. I also committed in babysteps so hopefully you can follow my process more easily :)

AlecAivazis:
Thanks for trimming this down! These changes look good to me. Looking forward to seeing the PR that makes queries fire in parallel

AlecAivazis:
I also appreciate that you tried to avoid any breaking changes by looking for `props` or `stuff` first. I think for the `v1` release we can make hooks return all props under the `props` key so we can better reflect kit's API. will leave a note in the discussion

PR #240, opened 2022-03-24
Partial results

### LINKED ISSUE #233: Are partial results already possible with Houdini?
Hi @AlecAivazis,

I just discovered this wonderful project. Endless congratulations.

I'm currently using Urql on all projects but as you may know Urql and Svelte don't get along very well (no SSR, broken (Svelte) API and others).

I haven't used Houdini yet (I'll try in the next few hours), but one thing that keeps me from changing is the following and from a quick read it doesn't seem to me that with Houdini you can get it today: **I HOPE I AM WRONG**.

This is what the world calls: `Partial results` which Urql obtains with [Schema awareness](https://formidable.com/open-source/urql/docs/graphcache/schema-awareness/).

This is one of the biggest points in favor of solutions like urql or apollo for me. And that's one of the reasons I really enjoy working with Graphcache (normalized data) and schema awareness. Although it makes everything much heavier (in terms of development and runtime).

The UX show in the hands of the user is that the application looks native and almost offline (if the internet is fast enough). The loading spinners disappear completely and everything seems immediate.

Ultimately that's why I chose to complicate my life by choosing to develop single page apps (SPAs) instead of multi-page (server rendered) apps (MPAs).

**A very simple example is the list > detail**. As soon as I enter the details I can show the data that I have already shown in the list and download in the background (almost without the person noticing, especially if I do it in the prefetch) the data that did not appear in the list.

I know that you understand what I mean and I also know that it is difficult to understand well how to organize this thing.

Is it already possible today with Houdini?

If you answer me yes, I could pass out with joy... :smile:

AlecAivazis:
This PR closes #233 and extends the `CachePolicy` behavior to accommodate partial results from the cache when the user opts into partial results with `@cache(partial: true)`. Along the way I added some more tests and improved support for writing null values over existing fields. These changes have been published under the `@next` version tag

Summary of new cache policy behavior in the presence of partial data:

- NetworkOnly doesn't need any changes
- CacheOrNetwork: if the data is totally available in cache then there's no network call. If the data is partial, the partial value is used and a network request will also be sent. If none of the data is present, a network request will be sent to look up the information. In some sense, Cache*Or*Network behaves like Cache*And*Network if there are partial results.
- CacheAndNetwork: Cache value will be used first regardless of partial or not and then a network request will be sent for the rest.

Changes:

- `cache.read` needs to return whether the result is a partial response or not which means the actual data needs to be nested inside of an object that has two fields for now: `results` and `partial`
- `fetchQuery` needs to return the partial state and pass it onto the query runtime. In order to make this simpler, i changed the way that the query results are threaded as props to a component. Instead of being multiple values, its now a single object with keys `result`, `variables`, `source`, `partial` (cc @fehnomenal)
- the `onLoad` handler for a query needs to check if the policy is `CacheOrNetwork` and has a `partial` result so that it can send a request to load the rest of the data
- removed `cache._internal_unstable.isDataAvailable` since it duplicated a lot of the same work as computing `partial`

Todo:

- [x] Reading an empty value for a non-null field needs to cascade up to the nearest nullable field
- [x] Test against real applications
- [x] Add documentation: https://github.com/HoudiniGraphql/docs/pull/4
- [x] Add `partial` field to return of `query` 
- [x] Make partial data opt-in (with configurable default)


frederikhors:
Amazing! You're a real genius!

Do we have a complex demo app to try?

frederikhors:
@jycouet do you know about this?

AlecAivazis:
@frederikhors - not yet. i'm going to try to get something public together this week to showcase the behavior. I've tested the changes against a few smaller apps so far but haven't found the time yet to test against a real-world application. Will let you know when there's something you can play with

frederikhors:
I think we need a demo app that slowly grows complex (and I can help too).

jycouet:
@frederikhors thank you for putting us in contact @AlecAivazis and me.
We are talking together now to see what could we do or not do together. (and it's very nice exchanges whatever the outcome will be).

We will share our thoughts in the discussion you opened somewhere else in the near future.

AlecAivazis:
Okay I found some time to test this out in a real example but its not open sourceable so I'm still going to try to get something out this weekend. @frederikhors if you want to give this a shot yourself, you can pull down `houdini@next` and `houdini-preprocess@next` and give it a shot. Feel free to reach out on discord if you want a more direct way of talking through it

frederikhors:
Il will try to create the same demo I created for kitql: https://github.com/jycouet/kitql/pull/94 ASAP.

AlecAivazis:
@frederikhors i have a question for you. When there _is_ partial data, what is the purpose of having a `partial` key in the response? It doesn't really tell you much so you wouldn't really be able to use it to do anything other than replace the entire contents with a loader. The user would still have to check all of the nullable fields in the response to see if they have a value

frederikhors:
> what is the purpose of having a partial key in the response?

Can you write an example?

I think I don't need it.

AlecAivazis:
In your initial ticket you said:

> And of course you need to have indicators like "isStale", "isFetching", something like that. So I can (if I want to) show indicators of running query.`

Im curious if you had a specific usecase in mind when you said that

AlecAivazis:
I've tested this against the pokedex example in a few different ways, including its pagination logic which gives me a good amount of confidence this is working as expected. I'm going to merge it so I can open up the optimistic updates PR i have waiting

PR #244, opened 2022-03-27
Generate type for usage in `afterLoad`

fehnomenal:
To simplify my typing I want to replace
```ts
  export function afterLoad({
    data,
    params,
  }: Parameters<Load>[0] & {
    data: {
      GetCatalogPublic: GetCatalogPublic$result;
      SearchProductsPublicByCatalog: SearchProductsPublicByCatalog$result;
    };
  }): ReturnType<Load> {
```
with
```ts
  export function afterLoad({
    data,
    params,
  }: Parameters<Load>[0] &
    GetCatalogPublic$afterLoad &
    SearchProductsPublicByCatalog$afterLoad): ReturnType<Load> {
```

---

The generated types are super simple and have the form of
```ts
type <Name>$afterLoad = {
  data: {
    <Name>: <Name>$result
  }
}
```

I tried to do a global type that is generic over `Operation<any, any>`but I failed to add the name of the type as a property to the data object. Probably I didn't understand template literal types correctly :sweat_smile:

### DIFF
--- a/packages/houdini/cmd/generators/typescript/index.ts
+++ b/packages/houdini/cmd/generators/typescript/index.ts
@@ -114,6 +114,7 @@ async function generateOperationTypeDefs(
 	// the name of the types we will define
 	const inputTypeName = `${definition.name!.value}$input`
 	const shapeTypeName = `${definition.name!.value}$result`
+	const afterLoadTypeName = `${definition.name!.value}$afterLoad`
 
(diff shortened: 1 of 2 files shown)


AlecAivazis:
Very nice! I think this type deserves a section in the `typescript` guide of the doc site. Mind adding a PR there when you get a chance?

PR #245, opened 2022-03-30
Write internal fragments and directives to a file

### LINKED ISSUE #234: Add local directives to schema when pulling from server
List directives appear to be removed when generating the schema using the `--pull schema` flag.

At first, I added them to my schema manually like so:

`directive @list(name: String!, connection: Boolean) on FIELD`

but this produced an error when generating unless I formatted the comments exactly the same way as in the schema transform:

```
"""
@list is used to mark a field for the runtime as a place to add or remove
entities in mutations
"""
directive @list(name: String!, connection: Boolean) on FIELD
```

According to the contributing guide, there is a [separate transform for lists](https://github.com/AlecAivazis/houdini/blob/main/CONTRIBUTING.md#internal-graphql-schema) but the link is broken.

Am I correct in assuming that the intended behavior is for list directives to remain in the generated schema, even when pulling the server schema? I'd like them to remain for local validation and it's a bit inconvenient having to manage the directives manually. 

Thank you so much for your work on this wonderful project!

AlecAivazis:
This PR adds a new generator to the pile that's responsible for writing a `.gql` file to a customizable location so that IDEs can pick up the definitions.

I also took the opportunity to finally remove the `mode` config value 

Associated doc PR: https://github.com/HoudiniGraphql/docs/pull/7

fixes #234 

cc @ronkeiser


ronkeiser:
I just tried this out and it works great :)

PR #248, opened 2022-04-01
Optimistic responses

AlecAivazis:
This MR adds the API for providing an optimistic response for a mutation. All of the actual work was done in #228 when I rewrote the cache's storage mechanism but it's been used by a number of people now so I feel more comfortable releasing this API.

Todo: 

- [ ] Document API


AlecAivazis:
Closing this since it's based on the partial data branch and I dont want to conflate the two

PR #266, opened 2022-04-09
Give `afterLoad` the real data

fehnomenal:
I suppose this happened during the partial data thing


AlecAivazis:
Darn. Thanks for catching this! You're right this happened during the partial data work. I've been trying to think of a way to write tests for the preprocessor logic but it's been a struggle - do you have any ideas?

AlecAivazis (approved):
Thanks for getting this together so quickly!

Issue #268, opened 2022-04-09
Parallel queries


fehnomenal:
I think I will need this for my project so I will start working on this rather soon. But before starting I want to coordinate the changes so you are happy @AlecAivazis :).

My first attempt (in #238) was to put everything in loops and higher order maps and stuff and awaiting with `Promise.all`. This would mean input gathering has to happen beforehand for all queries before this and error handling afterwards.

For a much smaller diff in the tests (and thus the generated code and probably also the code generator) we could just defer awaiting the `fetchQuery` promise(s) after all calls (and before the potential `afterLoad` call so the data can be passed to the hook).

What do you think?

AlecAivazis:
Hey @fehnomenal! What you originally had is totally fine - I just wanted to keep the changes separate since there was a lot of stuff in flight. I think the only thing that really matters is that the requests happen in an `await Promsie.all`. I'll leave it up to you for the best way to pull that off 👍

AlecAivazis:
Just a heads up, in order to address an issue I ran into with #63 I am going to need to merge `houdini` and `houdini-preprocess` into a single package which is something I've been meaning to do anyway. If you end up starting this before that gets merged, you will likely run into some conflicts. If you want to get your version out soon, I can wait to make my change and get your work in first. Just let me know how quickly you think you can submit the change.

fehnomenal:
Well we could just do something like
```ts
const _TestQuery2Promise = fetchQuery(...)
const _TestQuery1Promise = fetchQuery(...)

const _TestQuery2 = await _TestQuery2Promise
if (!_TestQuery2.result.data) {...}

const _TestQuery1 = await _TestQuery1Promise
if (!_TestQuery1.result.data) {...}

return {
  props: {...}
}
```

where I hope to keep the code diff rather small. The runtime behavior might be a bit different than doing `Promise.all` but it should be similar to what it is now so hopefully not a breaking change :sweat_smile: 

I will start working right now on this.

AlecAivazis:
Ah yea, good point. I'm okay with that too

AlecAivazis:
Closing this now that #269 is in - thanks again!

PR #269, opened 2022-04-10
Parallel queries

fehnomenal:
See #268 

I changed the approach to only split into storing promises and awaiting them if one `load` function has to execute more than one query.

I could not yet test it but from looking at the generated code there shouldn't be any problems.

### DIFF
--- a/.changeset/rotten-frogs-deny.md
+++ b/.changeset/rotten-frogs-deny.md
@@ -1,0 +1,5 @@
+---
+'houdini-preprocess': minor
+---
+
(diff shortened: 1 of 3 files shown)


AlecAivazis:
Thanks for getting this together so quickly. I'll find some time to test it out over the weekend so we can merge/deploy it before the custom ID work goes in

AlecAivazis (approved):
I modified the example to ask for two queries and the requests were sent in parallel. nice work!

image

PR #273, opened 2022-04-13
Consolidate Houdini modules under one package

### LINKED ISSUE #241: Add a warning if runtime was generated with a different version than the preprocessor
We need to warn the user if the preprocessor and generated runtime are using different versions. I just ran into this myself so it's very likely going to bite someone else 😅.

AlecAivazis:
This PR closes #241 by consolidating `houdini`, `houdini-preprocess` and `houdini-common` under a single `houdini` package in order to allow for more logic shared between the runtime and the other packages. This means that the preprocessor is now imported like so:

```javscript
import houdini from 'houdini/preprocess`
```

TODO:
- [x] Update getting started guide
- [x] Update links in contributing guide


AlecAivazis:
Pulled this down into the pokedex app and verified that everything worked as expected

PR #279, opened 2022-04-16
add pinst to disable postinstall on packaging

### LINKED ISSUE #278: don't run husky install on projects depending on houdini
Currently, the posinstall script tries to work even for projects that don't have husky installed. However, because I'm using houdini in a monorepo, where it's not installed in the main repository folder, and I am using husky, it breaks on this step with `Error: .git can't be found (see https://git.io/Jc3F9)` - which isn't caught.

This could be fixed just by adding more conditions to the error handler, but I think a better solution would be to move this step from `postinstall` to `prepare` (which is now suggested in [Husky docs](https://typicode.github.io/husky/#/?id=install), see https://github.com/typicode/husky/issues/884), as `prepare` scripts don't run on dependencies at all.

Edit: I just realized houdini is using yarn, and it stopped supporting `prepare` in v2. It appears the way Husky recommends to avoid pushing the script onto people depending on a package is to use [`pinst`](https://github.com/typicode/pinst) by the same author to remove postinstall on `prepack` hook: https://typicode.github.io/husky/#/?id=yarn-2
```json
"scripts": {
    "postinstall": "husky install",
    "prepack": "pinst --disable",
    "postpack": "pinst --enable"
  }
```

oplik0:
resolves #278 by implementing the method [husky docs suggest](https://typicode.github.io/husky/#/?id=yarn-2) to disable postinstall for publishing with [pinst](https://github.com/typicode/pinst).

### DIFF
--- a/.changeset/violet-shoes-hide.md
+++ b/.changeset/violet-shoes-hide.md
@@ -1,0 +1,5 @@
+---
+'houdini': patch
+---
+
+add pinst to disable postinstall for publishing
--- a/package.json
+++ b/package.json
@@ -33,7 +33,9 @@
 		"changeset": "changeset",
 		"version": "changeset version",
 		"release": "yarn build && changeset publish",
-		"postinstall": "node -e \"try { require('husky').install() } catch (e) {if (e.code !== 'MODULE_NOT_FOUND') throw e}\""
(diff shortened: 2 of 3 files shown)


AlecAivazis (approved):
Thanks for hunting down the correct fix!

PR #283, opened 2022-04-20
Given component queries better default prop values

### LINKED ISSUE #282: Querying in component tries to access key of undefined value
To reproduce create a component under `src/lib` and use it in `__layout.svelte` or any other route.

The generated code fails here:
```js
function instance$9($$self, $$props, $$invalidate) {
  let { _Search = void 0 } = $$props;
  let _Search_handler = query({
    "config": config,
    "initialValue": _Search.result, // <-- here
    "variables": _Search.variables,
    "partial": _Search.partial,
    "kind": "HoudiniQuery",
    "artifact": _SearchArtifact,
    "source": _Search.source
  });
```

Am I doing something wrong?

AlecAivazis:
This PR closes #282 by providing a default value of `{}` to the props driven by the query response

### DIFF
--- a/.changeset/polite-pans-hammer.md
+++ b/.changeset/polite-pans-hammer.md
@@ -2,4 +2,4 @@
 'houdini': patch
 ---
 
-Fixed bug in component queries
+Fixed bug in component queries associated with the unloaded response from `load`.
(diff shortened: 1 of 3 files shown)


fehnomenal (commented):
Works perfectly, good job :+1:

AlecAivazis:
Thanks for verifying @fehnomenal!

Issue #247, opened 2022-04-01
Pass cache policy to refetch


AlecAivazis:
It should be possible to specify a cache policy to use when triggering a refetch. This would be useful to users that want to use `refetch` instead of `@paginate` to navigate a long list

AlecAivazis:
This has been fixed as part of `0.15.0`

PR #291, opened 2022-04-25
✨ Store-based APIs (KitQL integration)

### LINKED ISSUE #189: Calling queries in endpoints
Its ofcourse possible to just use fetch. But I think it would be nice if there is a way of using houdini in server endpoints

### LINKED ISSUE #255: Supporting importing graphql documents from files
A user should have the freedom to decide if they want to import their documents from a file or define them inline.

jycouet:
This PR adds a new store-based API to houdini, very similar to what KitQL gives users. This PR was done by
@AlecAivazis and @jycouet to take advantage of approaches from both libraries.  

closes #255, closes #189

## ✅ Things to check / notes / comments

- [x] Client side navigation no await @jycouet  (with blocking param option)
  - [x] Add `blocking=true` parameter to `queryLoad` 
- [x] ~context and session should be optional~. Done differently with `event` on load side and that's it! @jycouet  
- [x] `query` & `queryLoad` => `fetch` @jycouet 
- [x] Integration tests à la playwrite 
- [x] Improve variable existence check in QueryStore @jycouet 
- [x] type definitions exported from root package
- [x] Preprocessor ensureImports bundles everything for every route 😢
- [x] Add warning when getHoudiniContext is called outside of component init @jycouet 
- [x] Operation description
  - [x] external `.graphql` files
  - [x] external `.gql` files @jycouet  https://github.com/HoudiniGraphql/houdini/issues/255
- [x]  `setEnvironment`  -> `houdiniClient.init()` @jycouet
- [x] Pull houdiniContext in paginated runtimes
- [x] Improve console log of generate command
  - [x] `level` management in logs maybe:
    - null => no logs at all (quiet? even super quiet!)
    - 0 => Just "It's done"
    - 1 => Smart logs
    - 2 => Module start / stop
    - 3 => Maxi detail
  - [x] smart logs are for 🥳
  - [x] Maybe instead of config `quiet: boolean` you have now: `logLevel: 'quiet' | 'smart' | 'full' = 'smart'`, what do you think?
- [x] Environment / kitqlClient
  - [x] Remove multiple client examples from houdini example 
  - [x] I wonder if instead of `new Environment(...)` we could have `new 
- [x] Typings
  - [x] Query variables
  - [x] Query result 
  - [x] Store type? `type ${storeName}_data`?
- [x] Load SSR / Load Client (CSR)
  - [x] load
  - [x] component queries
  - [x] rename `load` @jycouet 
- [x] Normalized caching
- [x] Normalized caching with multiple ids
- [x] Query, variable change, caching
- [x] Cache Policy / Overwrite (when we do a `.query()` / Default Behavior https://github.com/HoudiniGraphql/houdini/issues/247
- [x] Marshal
- [x] Store APIs
  - [x] Query
  - [x] Mutation 
  - [x] Fragment
  - [x] Subscription 
  - [x] Paginated Query
  - [x] Paginated Fragment
- [x] Update Preprocessor 
  - [x] Queries 
  - [x] Mutations
  - [x] Subscriptions 
  - [x] Fragments 
  - [x] Paginated Fragments
  - [x] Paginated Queries
- [x] Update runtime to use store APIs
  - [x] Route Query
  - [x] Component Query
  - [x] Paginated Query
  - [x] Fragment 
  - [x] Paginated Fragment
  - [x] Mutation
  - [x] Subscription  @jycouet 
- [x] Check `TODO` in the code

## To talk about?
- Speak about defaults in `npx init`

## Post Houdini publish
- Creating a `buildFetchQueryFn` in KitQL. @jycouet

## Will consider later
- ResetAllCaches 
- cacheMs
- Integration with the Guild's CodeGen
- Integration with the Guild Scalars package / Client Scalars
- A way to specify unblocking navigation on generated preprocessor

And more to come 🥳


AlecAivazis (approved):

Jytesh:
Awesome update!

endigma:
Is this documented yet?

AlecAivazis:
Not yet! There is [an issue](https://github.com/HoudiniGraphql/docs/issues/12) tracking the progress for this update's documentation. I have added all of the smaller tasks but still have to document the actual store APIs. I'm going to try to get _something_ out in the next few days but until then, the best place to see how things are done would be the [integration tests](https://github.com/HoudiniGraphql/houdini/tree/main/integration/src/routes).

AlecAivazis:
@endigma there is now documentation for all of the stores available here: https://docs-git-0150-aaivazis.vercel.app/api/welcome.

There is also a new guide that goes over the relationship between document stores and the inline API: https://docs-git-0150-aaivazis.vercel.app/guides/working-with-graphql

AlecAivazis:
This has now been released in version version `0.15.0-next.1` which is available as `houdini@next`. Let us know if you run into any problems!

vhscom:
> There is also a new guide that goes over the relationship between document stores and the inline API [...]

I saw example code inside a `load` function:

```ts
await ViewerProfileStore.fetch({ event })
```

Where `event` is a `LoadEvent` identifier.

Was this design to ensure `event.fetch` is used, or are things like `params` used as well? Not at keyboard current but curious given my natural tendency is to destructure only what's being used in the `load` param list to satisfy principle of least privilege.

jycouet:
We went back and forth on param names and scope. _Not that easy topic!_

Yes, we need `fetch` for sure. Sometimes we need `session` as well. `params` could be used in some scenarios also.
So, in the end, we decided to go with `event` as you saw.

As you wrote, it looks nice and clean 👍
```ts
await ViewerProfileStore.fetch({ event })
```

You can of course restructure just after like 👇
```ts
  const { fetch, session, params } = event
  // Something to do with restructured args
  await ViewerProfileStore.fetch({ event })
  return {};
}
```

Another way could be 👇

```ts
export async function load({ fetch, session, params }) {
  // Something to do with restructured args
  await ViewerProfileStore.fetch({ event: { fetch, session, params } })
  return {};
}
```

PR #292, opened 2022-04-26
Pass variables to `afterLoad`

fehnomenal:
I want to eliminate duplicate code when I need the query variables also at other places.


AlecAivazis (approved):
Good idea and straightforward implementation - nice!

PR #295, opened 2022-05-06
Fix params for component query variable functions

### LINKED ISSUE #188: Pass session to non-route load
I'm using svelte-kit and houdini (looking awesome I might add). 
BUt I'm having difficulty getting session state in the variables. 

I'm expecting something like this to work. 
```svelte
<script context="module" lang="ts">
    export function MyQueryVariables({ page, fetch, session, stuff }): MyQuery$input {
        return {
            ...session.state
        }
    }
</script>
```

But the function is only being called with page as a parameter.

fehnomenal:
Fixes #188

### DIFF
--- a/.changeset/green-dogs-applaud.md
+++ b/.changeset/green-dogs-applaud.md
@@ -1,0 +1,5 @@
+---
+'houdini': patch
+---
+
+Pass contents of page and session store to variable functions for component queries
--- a/src/preprocess/transforms/query.test.ts
+++ b/src/preprocess/transforms/query.test.ts
@@ -439,7 +439,6 @@ describe('query preprocessor', function () {
 			`import { houdiniConfig } from "$houdini";`
 		)
 		expect(doc.instance?.content).toMatchInlineSnapshot(`
-		import { page, session } from "$app/stores";
 		import { routeQuery, componentQuery, query } from "$houdini";
 		import _TestQueryArtifact from "$houdini/artifacts/TestQuery";
 		export let _TestQuery = {};
@@ -461,9 +460,7 @@ describe('query preprocessor', function () {
 		    config: houdiniConfig,
 		    artifact: _TestQueryArtifact,
 		    variableFunction: null,
-		    getProps: () => $$props,
-		    getPage: () => $page,
-		    getSession: () => $session
+		    getProps: () => $$props
 		});
 	`)
 	})
@@ -492,7 +489,6 @@ describe('query preprocessor', function () {
 			`import { houdiniConfig } from "$houdini";`
 		)
 		expect(doc.instance?.content).toMatchInlineSnapshot(`
-		import { page, session } from "$app/stores";
 		import { routeQuery, componentQuery, query } from "$houdini";
 		import _TestQueryArtifact from "$houdini/artifacts/TestQuery";
 		export let _TestQuery = {};
@@ -514,9 +510,7 @@ describe('query preprocessor', function () {
 		    config: houdiniConfig,
 		    artifact: _TestQueryArtifact,
 		    variableFunction: null,
-		    getProps: () => $$props,
-		    getPage: () => $page,
-		    getSession: () => $session
+		    getProps: () => $$props
 		});
 	`)
 	})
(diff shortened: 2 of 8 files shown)


fehnomenal:
I use the following typings because no existing type seems to be appropriate:

```ts
export function FooVariables({
    page,
  }: {
    page: { stuff: App.Stuff };
    session: App.Session;
  }): Foo$input {
```

.changeset/green-dogs-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+'houdini': minor
```
AlecAivazis, line 2: I believe this is supposed to be a patch, right? I don't see  aren't any API-breaking changes

src/preprocess/transforms/query.test.ts
```diff
@@ -211,8 +209,7 @@ describe('query preprocessor', function () {
 		    queryHandler: _TestQuery2_handler,
 		    config: houdiniConfig,
 		    artifact: _TestQuery2Artifact,
-		    variableFunction: null,
-		    getProps: () => $$props
```
AlecAivazis, line 215: is `getProps`  not needed for some reason?

fehnomenal: It is only used in `componentQuery`:

https://github.com/HoudiniGraphql/houdini/blob/f92fdfa8bf612db50073da22d5e50bd176786cf5/src/runtime/query.ts#L212-L239

The commit removes it only for calls to `routeQuery`

.changeset/green-dogs-applaud.md
fehnomenal: Yeah you're right, sorry

src/preprocess/transforms/query.test.ts
AlecAivazis: Oh, gotcha! Yea the function isn't used in the route queries, I was just leaving it in for simplicity

src/preprocess/transforms/query.test.ts
```diff
@@ -444,6 +439,7 @@ describe('query preprocessor', function () {
 			`import { houdiniConfig } from "$houdini";`
 		)
 		expect(doc.instance?.content).toMatchInlineSnapshot(`
+		import { page, session } from "$app/stores";
```
AlecAivazis, line 442: These imports aren't valid in sapper or vanilla svelte. Instead you should import `getSession` and friends from `adapter.mjs` like you see [here](https://github.com/HoudiniGraphql/houdini/blob/1cce6a6e0d52a89b41c904b5714685fcbfa5db74/src/runtime/query.ts#L22). I'm actually surprised you needed to add these since I don't see them actually being used in `componentQuery` 🤔

AlecAivazis (changes requested):

src/preprocess/transforms/query.test.ts
fehnomenal: Ah yeah you're right. I will expand the generated adapter. I used `page` and `session` a few lines below by passing the via the `getPage` and `getSession` functions.

fehnomenal:
Currently, I can only test the kit implementation but I think sapper should work too.

From the code I think the old vanilla cannot work as the function returned `{}` where a store is expected (svelte's `get` calls `store.subscribe()` internally) so this must have failed. Well the last commit adds empty stores to emulate more or less the expected behavior.

src/preprocess/utils/ensureImports.ts
```diff
@@ -6,13 +6,18 @@ import { Config } from '../../common'
 
 const AST = recast.types.builders
 
-export default function ensureImports(config: Config, body: Statement[], identifiers: string[]) {
+export default function ensureImports(
+	config: Config,
+	body: Statement[],
+	identifiers: string[],
+	sourceModule: string
```
AlecAivazis, line 13: making `$houdini` configurable makes sense but i think we can make `$houdini` the default value for `sourceModule` which saves us having to pass it in over and over

AlecAivazis (changes requested):
One small change and then I think this is good to go! Sorry for submitting these reviews in multiple pieces, it's been a hectic week for me

src/preprocess/utils/ensureImports.ts
fehnomenal: No worries, take your time :)

AlecAivazis (approved):
Looks good! Thanks for being patient

PR #296, opened 2022-05-06
Load data for component query when nothing was cached

fehnomenal:


AlecAivazis (approved):
Simple enough  - thanks for yet another contribution 👍

PR #309, opened 2022-06-03
Remove unused entry from lockfile

alexlafroscia:
When checking out this repo locally and installing the dependencies, I noticed that it causes a `yarn.lock` mutation to occur. This is probably not desired, so here's a PR to commit the change that `yarn` wants to make to the lockfile!

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -516,15 +516,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@babel/parser@npm:^7.13.4":
-  version: 7.13.4
-  resolution: "@babel/parser@npm:7.13.4"
-  bin:
-    parser: ./bin/babel-parser.js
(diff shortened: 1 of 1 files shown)


alexlafroscia:
I've never used the `changeset` tool before and the documentation doesn't mention what type of changes need one of those. If this feels like a change that doesn't require a changelog entry, I'm happy to remove that from the commit!

jycouet:
I would say that `changeset` are more for Houdini users, not really for developers.
Therefore I would remove the changeset entry.

I triggered the CI, but I will let @AlecAivazis have a look 👍

//Thanks for contributing//

alexlafroscia:
I went ahead and removed the changeset from this PR! I think that means CI will need to be kicked off again 🙃

AlecAivazis (approved):
Thanks for submitting this @alexlafroscia!

PR #310, opened 2022-06-03
Remove variables only used by internal directives

### LINKED ISSUE #308: Variables used only by directives should be removed from the request
I am running into an issue while using the `@append(parentID: ID)` directive as a hint to Houdini about how to correctly insert a newly-created record into a list

I have a query like this in my application, which is displayed on a route like `/list/:primaryKey`:

```gql
query ListPage($primaryKey: Int!) {
  list(primaryKey: $primaryKey) {
    id # `ID` for the purposes of implementing the `Node` interface
    primaryKey # Actual `id` in the database; queries use this data instead

    items @list(name: "ListPageItems") {
      id
      content
    }
}
```

Within that page, I have a button to create a new item within that list:

```gql
mutation AddItemToList($listPrimaryKey: Int!) {
  createItemInList(listPrimaryKey: $listPrimaryKey) {
    ...ListPageItems_insert
  }
}
```

This works fine on the first list page navigated to (say, `/list/1`) but breaks if I perform a navigation to another list (like `/list/2`) without fully reloading the page. If I stay on the initial list, the list is inserted into automatically. If I navigate to the second list and try to perform the mutation, the operation is successful but the item just doesn't appear in _either_ of the lists. Refreshing `/list/2` then shows the item correctly, since it's loaded anew from the API.

In an attempt to fix this, I tried to use the `@append` directive, updating my mutation like so:

```gql
mutation AddItemToList($listID: ID!, $listPrimaryKey: Int!) {
  createItemInList(listPrimaryKey: $listPrimaryKey) {
    ...ListPageItems_insert @append(parentID: $listID)
  }
}
```

However, this causes me to run into a new error: the server complains that `$listID` is never actually used in the mutation, and thus should not be provided.

This isn't wrong, from the perspective of the API, since the entire `@append` directive only exists on the client; as far as I can tell, the actual API call has the directive removed. However, I can't just remove the variable from the mutation either, since the client needs it.

I _think_ that any variables used only by a directive, and never used in the actual GraphQL query, should be stripped out of the query before sending it to the API. Does that sound like the right solution to this problem?

alexlafroscia:
This PR removes the variables only used by internal directives. See the related issue for more details about why this is important.

Closes #308


src/cmd/generators/artifacts/index.ts
```diff
@@ -103,6 +107,51 @@ export default async function artifactGenerator(config: Config, docs: CollectedG
...
 						},
+
+						VariableDefinition(variableDefinitionNode) {
+							const {
+								value: definedVariableName,
+							} = variableDefinitionNode.variable.name
+							let isOnlyUsedByInternalDirective = true
+
+							// Walk the document to find locations where the defined variable is used
+							graphql.visit(document, {
```
alexlafroscia, line 118: Calling the `.visit` API within another call to the `.visit` API doesn't feel great, but I don't think it can really be avoided; trying to keep track of all of the required state in a single pass of the AST might not be possible, since the `VariableDefinition` visitor will have completed before any of the `Argument` visitors have been called.

If there's a way to avoid this, please let me know!

src/cmd/generators/artifacts/artifacts.test.ts
```diff
@@ -265,6 +265,34 @@ test('internal directives are scrubbed', async function () {
...
+		'utf-8'
+	)
+	expect(queryContents).toBeTruthy()
+	// parse the contents
+	const parsedQuery: ProgramKind = recast.parse(queryContents, {
+		parser: typeScriptParser,
+	}).program
+
+	// verify contents
+	expect(parsedQuery).toMatchSnapshot()
```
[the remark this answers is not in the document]
alexlafroscia: I ended up copying another `toMatchInlineSnapshot` call from a neighboring test and manually modifying it to match the desired value.

AlecAivazis (commented):
Thanks for getting this out so quickly!

src/cmd/generators/artifacts/index.ts
AlecAivazis: Whoa, i'm not sure how my original comment got lost but I think we can avoid the nested walk if we just do a second pass after remove the internal directives. If we keep track of every variable that was actually used in the document we can remove the definitions that weren't used from the list of operation argument definitions.

alexlafroscia: Ah yeah, that's a good idea! I'll try refactoring it like that 👍

alexlafroscia:
@AlecAivazis is there a reason this project is using such an old version of TypeScript? I am running into a problem with the code I am refactoring to, where the TypeScript version provided by my editor's language tools are OK with the code (since it defaults to TypeScript 4.6) but the tests are failing because the version used by the project (3.9.5) doesn't seem to like it.

AlecAivazis:
Like I mentioned in the other PR, I am more than happy to update the typescript version if we can get the build to pass. Feel free to update it here instead of a separate PR if you prefer that way you don't have to juggle too many things going in at once

alexlafroscia:
@AlecAivazis gotcha! Juggling a bunch of branches doesn't matter to me; keeping things separate is fine by me.

In case I can't figure out why the TypeScript version bump breaks CI, feel free to merge this PR without the TypeScript update if that's what you want to do! I would love to get a Houdini version published with this bug fix in it so I can fix my app at work 😬

alexlafroscia:
Since I was able to get #312 passing, if you're happy with both PRs, let's merge that one first; I'll then rebase this one on `main` again to I can remove the silly `@ts-expect-error` comment that's currently necessary for this code to work correctly.

AlecAivazis:
@alexlafroscia its merged now! This should be good to go once we get the 👍 from github

AlecAivazis (approved):
Looks good! Thanks for adding a test 👍

PR #312, opened 2022-06-05
Remove Resolution to TypeScript 3.9.5

alexlafroscia:
I came across this old version of TypeScript while writing #310 and finding a case where something wouldn't type-check that should have.

Removing the resolution seems safe; the tests still pass, so I figure it's worth removing to use the latest version of TypeScript instead!

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
 		"ts-ast-serializer": "^1.0.1",
 		"ts-jest": "^26.0.0",
 		"ts-node": "^9.1.1",
+		"tslib": "^2.4.0",
 		"typescript": "^4.6.3"
 	},
 	"dependencies": {
--- a/package.json
+++ b/package.json
@@ -37,9 +37,6 @@
 		"prepack": "pinst --disable",
 		"postpack": "pinst --enable"
 	},
-	"resolutions": {
-		"typescript": "3.9.5"
-	},
 	"devDependencies": {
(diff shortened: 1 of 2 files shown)


AlecAivazis:
I'm pretty sure somewhere along the way I ran into a build error and could only fix it by downgrading. I'd definitely prefer to use the latest version if possible.

alexlafroscia:
That didn't end up being too bad to fix!

The complaint from the compiler was this:

> (plugin commonjs--resolver) Error: /Users/alexlafroscia/Code/github.com/HoudiniGraphql/houdini/src/common/graphql.ts(92,19): semantic error TS2807: This syntax requires an imported helper named '__spreadArray' with 3 parameters, which is not compatible with the one in 'tslib'. Consider upgrading your version of 'tslib'.

Since `tslib` wasn't installed as a direct dependency, TypeScript was left to guess which version to use. Because it's a dependency of many of our dependencies, it could find _a_ version, but not a _compatible_ version.

Just adding it to the dependencies for `houdini` gives it access to a version it can work with!

alexlafroscia:
One thing I'm curious about: it seems like it was choking on transpiling a spread operator, which I was surprised it needed to do at all. It seems like the compiler targets ES5 output. Is that necessary?

AlecAivazis (approved):
Ah ha! Nice catch 👍

PR #326, opened 2022-06-13
Mutation results shouldn't be nullable by default

AlecAivazis:
@alexlafroscia brought up a good point in #304 about the nullability of mutation results. This PR fixes the generated type for mutation artifacts to not be able to be `undefined`


alexlafroscia:
Thanks for this, @AlecAivazis !

PR #327, opened 2022-06-13
Multiple schema file - Add new test

jycouet:


src/cmd/main.ts
```diff
@@ -57,17 +58,28 @@ program
...
-						config.apiUrl,
-						config.schemaPath
-							? config.schemaPath
-							: path.resolve(targetPath, 'schema.json'),
-						args.pullHeader
-					)
-					console.log(`Pulled latest schema from ${config.apiUrl}`)
+
+					// if the schema path is a glob, tell the user this flag doesn't do anything
+					if (config.schemaPath && glob.hasMagic(config.schemaPath)) {
```
AlecAivazis, line 63: Something is weird about this diff. these changes are already present in the main branch: https://github.com/HoudiniGraphql/houdini/blob/main/src/cmd/main.ts#L63-L68

.changeset/thin-readers-matter.md
```diff
@@ -0,0 +1,5 @@
+---
```
AlecAivazis, line 1: I'm not sure changes in the integration tests need to bump the package version since there's nothing new for the user

PR #344, opened 2022-06-17
0.15.0 Final tweaks

AlecAivazis:
This PR hold all of the little tweaks that have been made while getting `0.15.0` ready. Releases are done manually to avoid the noise of 2 PRs per tweak.

Things to do:
- [x] fragment should only return a nullable value if the `initialValue` is also nullable
- [x] component queries
- [x] `this.error` from variable functions
- [x] Address double query 
- [x] fix session retrieval
- [x] scalar double marshal
- [x] subscriptions with overlapping CSF and load
- [x] pagination + scalars
- [x] pagination with abstract types
- [x] global config option to disable masking
- [x] message about updating source path when upgrading to 0.15.0
- [x] missing scalar definitions should generate `any`
- [x] notification about dropping sapper support


AlecAivazis:
Okay. The latest version has been published as `0.15.0-next.12` and we think it's pretty darn close to ready for people. 

If anyone has some time to pull this down into their projects and let us know if you run into anything, that'd be amazing 🙌 ❤️  A small warning: the link to the docs is wrong since it points to where it'll be when things are live . Here is the correct one: [link].(https://docs-git-0150-houdinigraphql.vercel.app/). Also, if you have previously tried out a prelease of `0.15.0`, you'll have to rename `prefetch` to just `fetch` (one function does it all now).
cc @pixelmund @alexlafroscia @fehnomenal @plunkettscott.

src/cmd/generators/typescript/types.ts
```diff
@@ -54,7 +54,10 @@ export function scalarPropertyValue(config: Config, target: graphql.GraphQLNamed
 				return AST.tsTypeReference(AST.identifier(config.scalars?.[target.name].type))
 			}
 
-			throw new Error('Could not convert scalar type: ' + target.toString())
```
AlecAivazis, line 57: Note to self: make it generate `any` instead of throwing an error

fehnomenal:
I found indeed a minor formatting problem:

image

This is printed in the browser (Firefox) console.

AlecAivazis:
@fehnomenal that should now be fixed in `-next.25`

jycouet (approved):
Great work 🎉

imran-baitham:
.

PR #346, opened 2022-06-20
Assign dependencies more correctly between prod and dev and add peers

fehnomenal:
I'm at 0.15.0-next.10 and using pnpm as package manager.

In its default configuration it enforces strict peer dependencies meaning that they have to be installed or it will abort with an error. And this is why I noticed that houdini's declared dependencies seem to be a bit strange.

Specifically pnpm complained:
```
devDependencies:
+ houdini 0.15.0-next.10

 ERR_PNPM_PEER_DEP_ISSUES  Unmet peer dependencies

.
└─┬ houdini
  └─┬ @rollup/plugin-replace
    ├── ✕ missing peer rollup@"^1.20.0 || ^2.0.0"
    └─┬ @rollup/pluginutils
      └── ✕ missing peer rollup@^1.20.0||^2.0.0
Peer dependencies that should be installed:
  rollup@">=1.20.0 <2.0.0 || >=2.0.0 <3.0.0"
```

And I wondered why houdini needs the rollup replace plugin at runtime. As it turns out it is only used during building in `rollup.config.js` (or did I miss something?). So I think it should be moved to `devDependencies`, right?

More dependencies that are probably not needed at runtime (lol, this is confusing. Here runtime means when generating artifacts and building the app, NOT in the generated runtime):
```
- @changesets/changelog-git
- @changesets/cli
- @sveltejs/kit (only used for type imports?)
- @types/mock-fs
```

But the generated runtime needs `@kitql/helper` and `svelte`, so these should also go into `peerDependencies`.

I tested the above with the following `.pnpmfile.cjs` and everything seems to work fine (generating artifacts and building the sveltekit app):
```js
module.exports = {
  hooks: {
    readPackage(pkg, ctx) {
      if (pkg.name === 'houdini') {
        pkg.devDependencies['@changesets/changelog-git'] = pkg.dependencies['@changesets/changelog-git'];
        pkg.devDependencies['@changesets/cli'] = pkg.dependencies['@changesets/cli'];
        pkg.devDependencies['@rollup/plugin-replace'] = pkg.dependencies['@rollup/plugin-replace'];
        pkg.devDependencies['@sveltejs/kit'] = pkg.dependencies['@sveltejs/kit'];
        pkg.devDependencies['@types/mock-fs'] = pkg.dependencies['@types/mock-fs'];

        pkg.devDependencies['@kitql/helper'] = pkg.dependencies['@kitql/helper'];
        pkg.devDependencies['svelte'] = pkg.dependencies['svelte'];
        pkg.peerDependencies['@kitql/helper'] = pkg.dependencies['@kitql/helper'];
        pkg.peerDependencies['svelte'] = pkg.dependencies['svelte'];

        delete pkg.dependencies['@changesets/changelog-git'];
        delete pkg.dependencies['@changesets/cli'];
        delete pkg.dependencies['@kitql/helper'];
        delete pkg.dependencies['@rollup/plugin-replace'];
        delete pkg.dependencies['@sveltejs/kit'];
        delete pkg.dependencies['@types/mock-fs'];
        delete pkg.dependencies['svelte'];
      }

      return pkg;
    },
  },
};
```


AlecAivazis:
Fantastic! This is one of my last remaining tasks before we merged and released `0.15.0`. Thanks for doing the work! I'll take a closer look tomorrow morning but it sounds like you are thinking about these relationships correctly 👍

AlecAivazis:
I'm going to point this to the final tweeks branch and fix the issues locally

jycouet:
Hi @fehnomenal, on the `final-tweaks` branch I had to remove `peerDependencies` to have integration tests working again (https://github.com/HoudiniGraphql/houdini/pull/344/commits/579ffda6c5a79ae0ef0fe494e399f860efb131e2).

A part from this, all your suggestions are in. 👌

Could you have a look and let us know if it's ok like this or if you have a missing part without `peerDependencies`?

AlecAivazis:
thanks for all of the helpful info @fehnomenal. I feel a lot more comfortable with pnpm now 👍

fehnomenal:
>  I suppose we could bundle the runtime but I personally use the unbundled version a lot while debugging or tweaking things

True :joy:

PR #364, opened 2022-07-05
Improve support for enums

### LINKED ISSUE #316: Make GraphQL Enums Available at Runtime
### Discussed in https://github.com/HoudiniGraphql/houdini/discussions/311

<sup>Originally posted by **alexlafroscia** June  3, 2022</sup>
What do you think about generating a JavaScript representation of each GraphQL enum in your schema, that can be imported for use by the client application?

I have a number of enums that are used as inputs to queries in my GraphQL schema. I can import a type definition for the enum from `$houdini`, but there is no runtime definition for this enum. This makes it a little harder than it should be to feed enums from your client into a GraphQL query as an argument.

For an example based on my application, I need to make a query based on an enum value. Assume `ApplicationType` is some `enum` in GraphQL

```gql
query ApplicationSummaryPage($appType: ApplicationType!) {
  application(appType: $appType) {
    ...Whatever
  }
}
```

To actually make this query, I need to do something like this:

```typescript
import type { ApplicationType } from '$houdini';

function validateApplicationType(appType: unknown): ApplicationType {
  // Do some validation to ensure that the input is actually a valid `ApplicationType`
  // My code needs to manually keep a list of valid `ApplicationType` options because the app can't import it from anywhere
}

export function ApplicationSummaryPageVariables(page) {
  return {
    appType: validateApplicationType('SomeEnumValue')
  }
}
```

This ends up being kind of awkward; technically Houdini knows what the valid `ApplicationType` values are, since they are part of the TypeScript definition, but I can't actually import those at runtime in any way.

I would love for `$houdini` to make two changes to the enum handling:

1. Actually export the TypeScript `enum` in the type definition; right now I can import the type, but the TypeScript language server complains that it should actually be exported. Here's an example from my `$houdini` directory:
    CleanShot 2022-06-03 at 15 11 09@2x
2. Create a representation of the `enum` that is actually import-able at _runtime_ as well, so that enum options flow from GraphQL Schema `->` Houdini `->` my application code

If this feels like a reasonable idea, I'd love to take a crack at implementing this change!</div>

AlecAivazis:
This PR adds a generator that creates a JavaScript representation for a project's enums. It also updates the typescript generator to refer to this central definition.

closes #316


jycouet (dismissed):
Perfect, it's working well 🎉
I added a small second test ✅

integration/src/routes/preprocess/query/variables-error.svelte
```diff
@@ -1,5 +1,5 @@
 <script context="module" lang="ts">
-  export function PreprocessorTestQueryErrorVariables({}: { params: { id?: string } }) {
+  export function PreprocessorTestQueryErrorVariables({ params }: { params: { id?: string } }) {
```
AlecAivazis, line 2: Hm, if that's the case we probably want to tweak the types to accommodate this. A route's variables could be entirely driven by session values

PR #366, opened 2022-07-05
Improve isRoute logic

### LINKED ISSUE #281: `isRoute` is rather naive
I like to colocate components to the only route where they are used, so I end up with something like `src/routes/__layout.svelte` which uses `src/routes/__layout/header/Searchbar.svelte`. `isRoute` does what it's supposed to do and determines (by checking the file name prefix for `src/routes`) `Searchbar` is a route and not a component.

My easy workaround is to put such components into `/lib` or whatever but I think `isRoute` should be more smart, i.e. also taking the svelte config into account (https://kit.svelte.dev/docs/configuration#files).

AlecAivazis:
This PR improves the logic for distinguishing a route from a component. Queries can now be inside of `_NonRouteComponent.svelte` and it will be treated as a component query instead of a route. 

fixes #281

### DIFF
--- a/src/common/config.ts
+++ b/src/common/config.ts
@@ -157,8 +157,8 @@ For more information, visit this link: https://www.houdinigraphql.com/guides/mig
 				: path.join(this.projectRoot, '$houdini')
 
 		// if the config file specified an isRoute, use that
-		if (configFile.isRoute) {
-			this.configIsRoute = configFile.isRoute
+		if (configFile.routes) {
+			this.configIsRoute = configFile.routes
 		}
 	}
 
@@ -644,7 +644,7 @@ export async function getConfig(extraConfig?: Partial<ConfigFile>): Promise<Conf
 	if (config.framework === 'kit') {
 		// only load the route config if the user didn't specify one explicitly
 		await _config.loadKitConfig({
-			isRoute: !config.isRoute,
+			isRoute: !config.routes,
 			configFilePath: config.frameworkConfigFile,
 		})
 	}
--- a/src/common/config.ts
+++ b/src/common/config.ts
@@ -171,13 +171,20 @@ For more information, visit this link: https://www.houdinigraphql.com/guides/mig
 			return false
 		}
 
+		// only consider filepaths in src/routes
+		if (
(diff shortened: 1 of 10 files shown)


jycouet (approved):

fehnomenal:
Hmm, maybe I have a (or more) misconception(s) about the `static` config value.

In my new project I need to use the static adapter (it is the frontend of a tauri app) and naturally I set `static: true` for SPA mode. This of course means no `load` is generated and the query is executed after mounting. Now, in my route `routes/contracts/[id=integer].svelte` I need to access the id param and I tried it like the "normal" way inside the `...Variables` function, which, [according to the docs](https://www.houdinigraphql.com/api/query/inline#query-variables), gets the same arguments as without `static: true`.

But I get this (outputted to the browser console):
image
Using the `page` store also does not work (Cannot reference store value inside <script context="module">).

After removing `static: true` everything works as expected, i.e. `...Variables` get the correct arguments and the query also loads.
Now I'm wondering whether `static` is even necessary.

src/common/config.ts
```diff
@@ -145,6 +152,68 @@ For more information, visit this link: https://www.houdinigraphql.com/guides/mig
...
+	// compute if a path points to a component query or not
+	isRoute(filepath: string): boolean {
+		// a vanilla svelte app is never considered in a route
+		if (this.framework === 'svelte' || this.static) {
+			return false
+		}
+
+		// only consider filepaths in src/routes
+		if (
+			!posixify(filepath).startsWith(posixify(path.join(this.projectRoot, 'src', 'routes')))
```
fehnomenal, line 171: Can we make this configurable? In my tauri app svelte is at `src-svelte`

AlecAivazis: Can definitely make it configurable 👍 

I'm trying to remember what the original issue was that the static argument tried to fix. I'll try to hunt it down but I suspect  that kit's behavior has changed given what you're describing. If the static adapter triggers load functions and everything like normal then i agree we don't need it any more.

AlecAivazis: this is now configurable with the `routesDir` parameter

PR #377, opened 2022-07-06
Adding tests for partial data

jycouet:
# Let's add tests for Partial data 
_and getting familiar with the concept_

- [x] tests added
  - [x] http://localhost:3000/stores/partial/partial_List
    - [x] 4 users with id & name
    - [x] link to detail page
  - [x] http://localhost:3000/stores/partial/partial_2
    - [ ] id, name should be displayed instantly ❌
    - [x] birthDate should be displayed when the result arrive (after 1000ms delay)
    
---

 It was not clear to me where to put `partial: true`.
 So I did put it _everywhere_: 
- `houdini.config.js`: `defaultPartial: true,`
- src/routes/stores/partial/Partial_List.gql with `@cache(partial: true)`
- src/routes/stores/partial/Partial_User.gql with `@cache(partial: true)`

But none did the trick that I was expected, "Having id and name displayed instantly on the client-side navigation"

### DIFF
--- a/integration/houdini.config.js
+++ b/integration/houdini.config.js
@@ -6,11 +6,15 @@ const config = {
   module: 'esm',
   apiUrl: 'http://localhost:4000/graphql',
   defaultCachePolicy: 'CacheOrNetwork',
+  defaultPartial: true,
   scalars: {
     DateTime: {
       type: 'Date',
       // turn the api's response into that type
       unmarshal(val) {
+        if (val === null) {
+          return null;
+        }
         return new Date(val);
       },
       // turn the value into something the API can use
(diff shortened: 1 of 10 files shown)


integration/src/routes/stores/partial/Partial_List.gql
```diff
@@ -0,0 +1,6 @@
+query Partial_List {
+  usersList(snapshot: "Partial") @cache(partial: true) {
```
AlecAivazis, line 2: `@cache` needs to go on the query definition

integration/houdini.config.js
```diff
@@ -6,12 +6,16 @@ const config = {
...
   apiUrl: 'http://localhost:4000/graphql',
   defaultCachePolicy: 'CacheOrNetwork',
+  defaultPartial: true,
   routes: (filepath) => filepath && !filepath.includes('_') && !filepath.includes('+'),
   scalars: {
     DateTime: {
       type: 'Date',
       // turn the api's response into that type
       unmarshal(val) {
+        if (val === null) {
```
AlecAivazis, line 16: We should bake this into the unmarshal selection utility.

AlecAivazis (approved):

PR #381, opened 2022-07-07
✨ Improve onborading

jycouet:
The goal is to improve the onboarding experience.

1/ Adding a TypeScript option
2/ Adding the option to write everything or to log what is needed to finish (Aggressive / Less aggressive)
3/ Option for KitQL user to have graphql endpoint in kit
4/ default esm => do not write
5/ default kit => do not write
6/ info about Sapper (deprecated)
7/ remove the schemaPath option. (If they want to change, they will do after?)

What do you think? :)

### DIFF
--- a/src/cmd/init.ts
+++ b/src/cmd/init.ts
@@ -3,6 +3,7 @@ import inquirer from 'inquirer'
 import path from 'path'
 import { Config, getConfig, LogLevel } from '../common'
 import generate from './generate'
+import { createFolderIfNotExists, writeFile } from './utils'
 import { writeSchema } from './utils/writeSchema'
 
 // the init command is responsible for scaffolding a few files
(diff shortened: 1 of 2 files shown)


david-plugge:
Definitely some nice additions!

These are my ideas from #382:

- setup preprocessor
- setup @kitql/vite-plugin-watch-and-run
- setup [alias](@kitql/vite-plugin-watch-and-run)
- scripts
	- add "gen"
	- update "prepare" to also run "gen"
- add devDependencies
	- houdini
	- @kitql/vite-plugin-watch-and-run

AlecAivazis:
After talking with @jycouet a bit and reading the feedback from @david-plugge we're going to take this PR in a different direction and try to be a bit smarter about what we do. Instead of asking if they are using typescript we should look to see if typescript is a dependency of the project (or maybe its enough to just look for a `tsconfig.json`). 

I'm going to close this PR while we work on the new init but thanks again for the feedback @david-plugge

PR #405, opened 2022-07-12
Support query stores in endpoints

### LINKED ISSUE #402: Support query/mutation in endpoint
See: https://github.com/jycouet/kitql/issues/174

AlecAivazis:
This PR adds the necessary bits to support using a query store inside of a endpoint. the biggest difference between an endpoint and the other contexts is that `fetch` must be provided by the user explicitly (since we don't want to assume node fetch)

closes #402


jycouet (approved):
_Of course I need to build to run tests!!! 😅_

Very nice, let's go 🎩

liweihan:
Wow thank you @AlecAivazis for such a fast move!

PR #407, opened 2022-07-12
:construction: NEW: style 1, 2, 3, 4

jycouet:
Just to explore ideas


jycouet:
Style 4 SSR is not working because the preprocessor is not using the same `factory`.

---

Let's take an example, in `$houdini/stores/User.js` we export:
```js
// 1
export const GQL_User = factory()

// 2
export const UserStore = factory

// 3
export default GQL_User
```

Export 1 enables to be used everywhere (with the same factory)

```ts
import { GQL_User } from '$houdini'
GQL_User.fetch()
```

<br/>

Export 2 enables to use the same query with different parameters

```ts
import { UserStore } from '$houdini'
const myFactoryUserStoreForX = UserStore();
myFactoryUserStoreForX.fetch({ variables: { special: "X" }})

const myFactoryUserStoreForY = UserStore();
myFactoryUserStoreForY.fetch({ variables: { special: "Y" }})
```

<br/>

Export 3 enables to use the name you want for the existing `GQL_User` factory:
_Note: You have to configure `tsconfig.json` accordingly to get to `$houdini/stores/User`_

```ts
// file A
import MySuPeRcOoLnAmEoFuSeR from '$houdini/stores/User'
MySuPeRcOoLnAmEoFuSeR.fetch()

// file B
import a from '$houdini/stores/User'
a.fetch()
```

---
## Outcome

- I'm not sure we should support `export 3`. #Change_My_Mind ^^
- Preprocessor should probably use Export 1 to be able to mix styles.

jycouet:
@AlecAivazis, it's probably not the perfect place to speak about this.
Let's speak about it 👍

PR #423, opened 2022-07-18
Fix `stuff` bug in component query variables

### LINKED ISSUE #421: Wrong variable function input for inline component queries
I wrote about this already here: https://github.com/HoudiniGraphql/houdini/pull/366#issuecomment-1175407910 but now the problem resurfaced while trying to upgrade another project to the new store API.

Basically the problem is that the variable function only gets two values when being executed in the client, namely `props` and `session`:
https://github.com/HoudiniGraphql/houdini/blob/8be5953ae4237ef0f84346c595446ba8cd3feaee/src/preprocess/transforms/query.ts#L735-L764

My old code is using `stuff` via `page` but this is obviously no longer working.

Directly using `$page.stuff...` does not work because "Cannot reference store value inside <script context="module">".

Route `...Variable` functions get a `LoadEvent` passed. Can we try to pass component `...Variable` functions a trimmed `LoadEvent`? Something like:

```ts
type TrimmedLoadEvent = {
	params: Record<string, string>;
	props: Record<string, string>;
	routeId: string | null;
	session: App.Session;
	stuff: Partial<App.Stuff>;
	url: URL;
}
```

AlecAivazis:
This PR fixes a bug with the values passed to the variable function for component queries

closes #421

### DIFF
--- a/.changeset/great-rocks-wait.md
+++ b/.changeset/great-rocks-wait.md
@@ -1,0 +1,5 @@
+---
+'houdini': patch
+---
+
+fix bug when computing variables in component queries
--- a/src/preprocess/transforms/query.test.ts
+++ b/src/preprocess/transforms/query.test.ts
@@ -598,7 +598,9 @@ describe('query preprocessor', function () {
 		            prop4: prop4
 		        },
 
-		        session: _houdini_context_generated_DONT_USE.session
+		        session: _houdini_context_generated_DONT_USE.session(),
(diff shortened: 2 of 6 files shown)


fehnomenal (approved):

fehnomenal:
Works perfectly, thank you

AlecAivazis:
Thanks for checking!

PR #434, opened 2022-07-26
Prevent store data leaking across requests

AlecAivazis:
The previous store implementation created stores that existed at global scope and were shared across requests. This creates a risk that store data could be be shown to the incorrect user. This PR fixes this issue by updating query, fragment, and mutation stores to hold onto their data in a session-specific way, preventing any data from leaking to the wrong user. 

Note: there are no new tests since this is an entirely internal refactor. To verify these changes work, I'm going to increase the number of workers on the CI to try to create many concurrent requests.


AlecAivazis:
I had to bring the worker count down to 5 on github actions but i can consistently run the test suite with 20 workers on my machine.

integration/src/lib/utils/testsHelper.ts
```diff
@@ -59,29 +59,54 @@ export async function expectNGraphQLResponse(
   n: number,
   action: 'click' | 'hover' = 'click'
 ) {
-  // let nbRequest = 0;
+  // we are going to wait for n responses or 10seconds (whichever  comes first)
```
jycouet, line 62: I'm sure there is a good reason for this change. To be more consistent? Or?

I think 1 thing is missing now: 
- if we wait for 2 responses and we got 3, we should fail. 

What do you think?

src/runtime/lib/session.ts
```diff
@@ -0,0 +1,60 @@
...
+	// @ts-ignore
+	// get the reqID from the session
+	let { __houdini_session_key: reqID }: { __houdini_session_key: string } = session ?? {}
+
+	// if we already have a reqID, use it
+	if (reqID) {
+		return reqID
+	}
+
+	// make sure that reqID isn't currently being used
```
jycouet, line 47: Good catch! I wouldn't think about this ^^ 👍
I was thinking using a generator uuid or v4, ... But math.random is way better!!! 😁

integration/package.json
```diff
@@ -13,7 +13,8 @@
     "package": "svelte-kit package",
     "previewWeb": "cross-env TZ=utc vite preview --port 3007",
     "preview": "npm run generate && concurrently \"yarn previewWeb\" \"yarn api\" -n \"web,api\" -c \"green,magenta\"",
-    "tests": "playwright test",
+    "tests": "playwright test --project default",
```
jycouet, line 16: How can you run only 1 test?
Before I was doing something like: 
`yarn workspace integration tests network`

jycouet (commented):
Looks great!
All tests are GREEN!!! Thx a lot.

I'm discovering `@replayio/playwright`, looks fantastic too 👍
What's the idea behind Playwright "projects"?

AlecAivazis:
> What's the idea behind Playwright "projects"?

They allow us to run the same test suite with different configurations. I was using them here to give us the option to run the test suite against a headless browser or replay but i think it'd be better to just have 2 different config files so we don't have to keep juggling flags

integration/package.json
AlecAivazis: This should be fixed now 👍

integration/src/lib/utils/testsHelper.ts
AlecAivazis: I mostly made this change for speed reasons. The old implementation added 1 second for every expect which meant we were waiting even if the responses came in the first 200ms. It also improved the consistency for larger worker pools since we could increase the maximum time to something long like 10s. 

> if we wait for 2 responses and we got 3, we should fail.

This is the down side of this new implementation. Waiting to see if we get a 3rd request means putting in a minimum time to wait. It just occured to me that the current implementation doesn't work when we  don't want _any_ requests so I'll fix that today 👍

jasonLaster:
Hey @AlecAivazis! Happy to answer any questions you might have. Also as a heads up **Replay Chromium** does not currently have sourcemaps. It will next week though. In the interim, we recommend **Replay Firefox**, but totally up to you. Also, if you'd like to create a new Houdini Playwright team, we can enable it for Test Suites so you can get the new Test Suite dashboard and PR comments.

We'll be announcing the dashboard in tomorrow's changelog, but here's a sneak peek https://www.notion.so/replayio/Changelog-37-Test-Suite-Dashboard-3ad6be27f04b4cb4982ea28811cc2322

AlecAivazis:
Really appreciate all of this support @jasonLaster! I was genuinely shocked to see that Jaril had forked houdini to get things working ❤️

Issue #433, opened 2022-07-23
Your project defines a Node interface but it does not conform to the Global Identification Spec.


Rar9:
Using drupal 9 graphql v4

Your project defines a Node interface but it does not conform to the Global Identification Spec.

If you are trying to provide the Node interface and its field, they must look like the following:

interface Node {
        id: ID!
}

extend type Query {
        node(id: ID!): Node
}

jycouet:
Strange to see all these servers having 
```graphql
type XXX implements Node
```
Not having `node(id: ID!): Node` implementation!

So, the server is saying, "Hey, you can use node(...) to query the type XXX" and actually, you can't!
Two options for endpoint builder:
1/ Don't `... implements Node` if you don't provide the implementation!
_or_
2/ Provide the implementation!

**With this kind of thing, `#GraphQL` has a high entry barrier!**

---

When the server tells Houdini _(by the introspection query)_ "hey, we have some types that implement Node", then Houdini will make use of it for you in some cases _(pagination for example)_!

Today, Houdini says that you have a problem with the schema _(and it's true, you have a problem with your schema!)_
![image](https://user-images.githubusercontent.com/5312607/180614342-8a3545b7-a614-4687-84fa-d702452e36aa.png)
And you should fix your schema (for reasons I explained above).

In the meanwhile, Houdini users are blocked 😓
I propose to:
1/ Switch the error to a warning (saying the same thing), and continue the normal flow
2/ Each time Houdini is using a Node implementation with a server not well configured => Throw the error msg.

jycouet:
For info, [Vendure](https://www.vendure.io/) will fix it here: https://github.com/vendure-ecommerce/vendure/issues/1666

AlecAivazis:
> 1/ Switch the error to a warning (saying the same thing), and continue the normal flow
> 2/ Each time Houdini is using a Node implementation with a server not well configured => Throw the error msg.

Both of these sound good to me. I think (2) might already happen 🤔

yasserlens:
> For info, [Vendure](https://www.vendure.io/) will fix it here: [vendure-ecommerce/vendure#1666](https://github.com/vendure-ecommerce/vendure/issues/1666)

For Vendure - what would you suggest we do until this fix is in place?
I still gete ❌ There is a problem with your project's schema error. Using Vendure version 1.6.4.

P.S: I'm new to GraphQL

Thanks!

jycouet:
[...]

PR #443, opened 2022-07-30
Node interface - warning if not properly defined && error on usage

jycouet:
Linked to https://github.com/HoudiniGraphql/houdini/issues/433

---

- [x] Depending on `logLevel`, the warning displays more or less info.

- [x] If your Node interface is not properly defined, but you don't use it:
image

- [x] If your Node interface is not properly defined, and you try to use it explicitly:

- [x] If your Node interface is not properly defined, and you try to use it implicitly (paginate for example):

- [x] as as before, with with `logLevel: full`

### DIFF
--- a/src/cmd/transforms/paginate.ts
+++ b/src/cmd/transforms/paginate.ts
@@ -5,6 +5,7 @@ import { ArtifactKind } from '../../runtime/lib/types'
 // locals
 import { CollectedGraphQLDocument, RefetchUpdateMode } from '../types'
 import { unwrapType, wrapType } from '../utils'
+import { getAndVerifyNodeInterface } from '../validators/typeCheck'
 
 // the paginate transform is responsible for preparing a fragment marked for pagination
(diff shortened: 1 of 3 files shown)


src/cmd/transforms/paginate.ts
```diff
@@ -288,7 +289,7 @@ export default async function paginate(
 			// figure out the 'target' type of the refetch
 			let targetType = config.schema.getQueryType()?.name || ''
 			if (fragment) {
-				const nodeInterface = config.schema.getType('Node') as graphql.GraphQLInterfaceType
+				const nodeInterface = getAndVerifyNodeInterface(config)
```
AlecAivazis, line 292: I dont think we want the verification here. This doesn't assume we need a node interface so the warning might confuse people

AlecAivazis (changes requested):
Small tweak and then I think we're good to go! 👍

AlecAivazis (approved):

Issue #424, opened 2022-07-18
Add way to toggle between isolated and global stores


AlecAivazis:
In order to support situations where a user wants to SSR a component query (or for some other reason interact with a store), we need to support a distinction between "isolated" and "global" stores. At the moment, all routes gets a local instance of a store but this clearly doesn't work for preloading store values on the server. 

```javascript
// 1
export const GQL_User = factory()

// 2
export const UserStore = factory

// 3
export default GQL_User
```

jycouet:
I would like to write the doc for this feature first. I think it can help elaborate.
Then we can have some reviews from different people and decide how to move forward?
In the end, as we will have opt-in/out at a different level, we will have great flexibility. That's why I think that the doc will be key.

_Of course, things will evolve from this original attempt! 😅_

---

# Store Modes

Stores have two modes: `Isolated` or `Global`, and you can configure the mode of each store at different levels: 
1/ Globally in houdini.config.js that will apply to all your stores.
```
{
  storeMode: 'Isolated' | 'Global'
}
```

2/ On each individual store, you can overwrite the behavior and specify the mode
```graphql
query MyQuery @Isolated | @Global { # @config(mode= 'Isolated' | 'Global')  ???
  id
  name
}

```

## Isolated mode
This mode is designed to have local stores, your data will be contained in a defined space.

A typical use case:
You have a component `<Users>` having one query `getUsers` with some args like `old: Boolean`
To display the two components at the same time, you want your mode to be `isolated` so that they are independent.

```svelte
<Users old={true}>
<Users old={false}>
```

Different syntax can ba used: 

```svelte
<!-- preprocessor -->
<script>
  import { graphql, query } from '$houdini'
  
  const { data } = query(graphql`query getUsers ( ... ){ ... }`)
</script>

{JSON.strignify($data, null, 2)}
```

```svelte
<!-- inline Document -->
<script>
  import { graphql } from '$houdini'

  const store = graphql`query getUsers ( ... ){ ... }`
    
  $: browser && store.fetch()
</script>

{JSON.strignify($store.data, null, 2)}
```

```svelte
<!-- external document                 query getUsers ( ... ){ ... } -->
<script>
  import { getUsersStore } from '$houdini'
  
  const store = getUsersStore()
  
  $: browser && store.fetch()
</script>

{JSON.strignify($store.data, null, 2)}
```

Houdini is exporting isolated stores like this:
```ts
// 1 => Not available in Isolated mode
[...]

PR #447, opened 2022-08-01
✨ New: store mode

jycouet:
Following: https://github.com/HoudiniGraphql/houdini/issues/424

- [x] 2 new configs:
  - [x] storeMode (default `Global`)
  - [x] storeModeInlineComponent (default `Isolated`)
- [x] new directive: `@houdini`
  - [x] With arg: `storeMode`, where you can switch mode query by query
- [ ] update doc 

---

@AlecAivazis could you check 2 things:
- I had some issues with `isRoute` (in generate or in preprocessor it was not giving me the same result! (because of absolute & relative path)). I fixed it and I added some tests, but you could probably double check 👍
- I created an `extractInfo` to avoid duplicating the logic, because I need to know is different places the storeMode. I thought that I will set it once in `artifact` and use it everywhere... But in the end it didn't seems easy to do it like this. (or a few refactoring will be needed). Like this it's working (I'm confident with what I tested :)) Let me know your thoughts on this point.


jycouet:
I'll close the PR for now.
Everything works on my end, nothing in github......................................................................................................................................... GREAT!

Issue #349, opened 2022-06-20
Add a directive to disable masking


AlecAivazis:
When migrating to houdini, users need a way to disable masking for a specific fragment mix as well as globally

AlecAivazis:
This can be done globally with the [disableMasking](https://www.houdinigraphql.com/api/config) option so it's no longer a priority

PR #449, opened 2022-08-02
Support SvelteKit's new API 

### LINKED ISSUE #242: Enforce route variable function if there are non-null arguments
A query like the following should not _require_ a variable function

```graphql
query AboutQuery($first: Int, $after: String, $last: Int, $before: String { 
   ...
}
```

### LINKED ISSUE #325: Preprocessor with external files
At the moment, if a user wants to use the preprocessor to generate a load function they must define their queries inline and use the appropriate "inline" function for the runtime. It would be nice to be able to somehow tell the preprocessor to generate a load function for a store. 

An example api: 

```svelte
<script context="module">
    import ViewerProfileStore from '$houdini/stores/ViewerProfile'

    export const routeStore = ViewerProfileStore
</script>

{$ViewerProfileStore.data.viewerProfile.firstName}
```

### LINKED ISSUE #348: Only regenerate runtime with new versions
Now that we have `meta.json` we can only update the runtime when we have a new version

### LINKED ISSUE #353: Generate type definitions for inline document variable functions
It's not very pleasant to type a variable function. `params` comes from `./_types/...` but the result is `foo$input`. We should take some inspiration from kit and generate some more helpful types

### LINKED ISSUE #395: Preprocessor configFile relative path
In a monorepo i have a project at `/packages/shop`.
To get the preprocessor working i have to set the relative path from the monorepos root directory. If the path changes i have to update the config file too.
```js
houdini({
  configFile: 'packages/shop/houdini.config.js'
})
```
This is working for me aswell, but it is somewhat ugly and i will always have to look it up for new projects
```js
houdini({
  configFile: new URL(import.meta.url).pathname.replace('svelte.config', 'houdini.config')
})
```

Beeing able to just put `./houdini.config.js` in there would be pretty nice.
Or even just a `.` and resolve the config from the current directory with its default name.
When working in a monorepo it could also check for a config file in the root directory and if there is none look up the current folder.

### LINKED ISSUE #398: Lazy inline queries
With the store api lazy queries are now easily possible by defining them in a separate file. But I would love to be able to define them inline.

My idea to implement this is:
We can probably detect whether the `query` call is part of a variable declaration or standalone, right? I.e. we can differentiate between
```ts
const { data, refetch } = query(...);
```
and
```ts
query(...);
```

If this is the case we could abort the preprocessor logic early to not execute the query in `load` or on mount.

My thinking is that in the second case the user clearly does not intend to use the query data right now. So it seems logical to defer execution.

What do you think, @aivazis @jycouet?

### LINKED ISSUE #428: Add pullHeaders to houdini.config.js
I´m using supabase graphql as my backend and every request needs an apikey header, even for pulling the schema.
In my case the apikey is set in an .env file and this is my current solution:
```json
{
  "pull-and-gen": "houdini generate --pull-schema -ph apikey=$(grep \"^VITE_SUPABASE_ANON_KEY=\" .env | cut -d \"=\" -f2)",
}
```
It would be nice to have access to the environment variables in the houdini config while houdini is pulling the schema. Since the config is also send to the browser where `process.env` is undefined a function can be used:

```js
const config = {
  ...

  pullHeaders() {
    return {
      apikey: process.env.VITE_SUPABASE_ANON_KEY
      // or maybe use 'import.meta.env' instead
    }
  }
};
```

_Edit:_

Same with the apiUrl, that is also defined as an env variable.

### LINKED ISSUE #452: Accessing GQL store while loading in an other route breaks SSR
Demo: https://github.com/janvotava/houdini-error-reproduce-001

When the `$GQL` store is read in the [__layout.svelte](https://github.com/janvotava/houdini-error-reproduce-001/blob/main/src/routes/__layout.svelte#L12), the resulting data in the [routes/index.svelte](https://github.com/janvotava/houdini-error-reproduce-001/blob/main/src/routes/index.svelte#L21) will be `null` for some reason and SSR won't work.

### LINKED ISSUE #460: Customizing prefix to have more options than only "GQL_"
additionally, customizing `GQL_` would be nice, some projects have case conventions.

_Originally posted by @endigma in https://github.com/HoudiniGraphql/houdini/issues/391#issuecomment-1206668510_

AlecAivazis:
This PR updates houdini's APIs to be compatible with the updates to SvelteKit as described here: https://github.com/sveltejs/kit/discussions/5774. 

## Update 8/21

Almost there! 🎉 

All of the integration tests are passing locally but there seems to be something going on with GitHub Actions.  Hopefully it doesn't take us too long to figure out what's going on. 

For an up-to-date overview of what's changed, you can check out the [release notes](https://docs-git-directory-routes-houdinigraphql.vercel.app/guides/release-notes) section of the doc's preview link. The content below is all still valid but I'm not going to nitpick small changes in this PR description so it's best to just go to the updated docs directly.

Thanks for the patience!

## Upcoming Changes

Just to get some points out of the way for people worried about major breaking changes:
- SvelteKit projects will now need to use the vite plugin defined at `houdini/vite` which replaces the old preprocessor (and the ugly `server.fs.allow` config). But don't worry, there are **lots** of really cool things that we're getting with this move.
- The external document store api for SvelteKit routes is changing. This means that if you rely heavily on the document stores, you will have to update your load functions but as a result you can delete a lot of the code in your components and the overall mental model is a lot simpler (no more passing around `context`)
- inline queries have to change too (since kit just made `data` a keyword in a route).  At a high level, you'll now use the result of `query` directly instead of destructuring data.  Also, everything that used to be defined in `context="module"` is now defined in `+page.js/ts`.

For additional breaking changes and notes, check out the section at the bottom of this. We know this is a lot of changes but we have been thinking about some of these for awhile and since SvelteKit has forced us to change the API, it seemed like a good time to shift some things around.

## Vite Plugin

For kit projects, the preprocessor has moved to `houdini/vite` and should be included as a plugin in your `vite.config.js`. You should also remove the the `fs.server.allow` config since its baked into the plugin.  This plugin not only processes our components and javascript files, but it also integrates into vite's `dev` script to bring some huge improvements:

- automatically generates your artifacts and stores whenever it detects a change in your source code. if you have configured your own watcher, it might not be necessary anymore.
- poll your api for changes in your schema and regenerate when they are detected. This behavior is customizable between 3 modes: a duration in milliseconds, only pull when the project first starts, and never pull.

Ideally, you shouldn't need to run `generate` ever again! It'll even get run when you build your application with `vite build` so feel free to delete that `generate` script from your `package.json` if you have it.

## New Store API

The core of this change is that instead of using stores that you import, your component will now get the store from its props. This is the full example for the component-side of a store-driven route:

```svelte
<script>
  /** @type {import('./$types').PageData */
  export let data: PageData ;

  $: ({ UserInfo } = data);
</script>

{$UserInfo.data.firstName}
```

Notice there's no more need to thread variables around or that ugly `$: browser && store.fetch`.

In order to make this work, your `load` functions need to now return instances of the store embedded in the correct key. This can be easily done with the new `load_StoreName` functions. The `load_` prefix is to make it easy to autocomplete when you are working in your editor:

```javascript
import { load_UserInfo } from '$houdini'

    return { 
        ...await load_UserInfo({ event })
    }
}
```

This means that when adding a query to a route, tasks like adding a store to a load, pulling that store out of data in your component, etc can be auto completed from just starting with `load_` in your route's load function. It does make the manual logic that uses a fetch result kind of messy since they have to pull out UserInfo but I think that’s a rare situation so I’m in favor of supporting the most people with the clean API and letting the advanced users deal with the complexity.

Also, on a slightly unrelated note: you don't need to pass `context` everywhere anymore! that all happens behind the scenes now.

## Page Queries

You can now define a `+page.gql` file inside of a route directory to automatically opt-into a generated load without having to do anything else:

```graphql
# src/routes/myProfile/+page.gql

query MyQuery { 
	viewer { 
		id
	}
}
```

With that file in place, you can just import and use the `MyQuery`  store passed to your route and the view will be SSR'd automatically.

## Generating Loads For Stores

You can now tell the houdini plugin to generate loads for your stores. To do this, you need to export a `houdini_load` variable from your `+page.js/ts`  file:

```jsx
// src/routes/myProfile/+page.ts
import { GQL_MyQuery, GQL_Query1, GQL_Query2 } from '$houdini'

export const houdini_load = GQL_MyQuery 
// or 
export const houdini_load = [ GQL_Query1, GQL_Query2 ]
```

This can be mixed with the new `graphql` tag api to define your queries inside of your javascript files:

```typescript
import { GQL_MyQuery, graphql. } from '$houdini'

const otherQuery = graphql`
	query ViewerInfo { 
		viewer { 
			id
		}
	}
`

export const houdini_load = [ GQL_MyQuery, otherQuery ]
```
or 

```typescript
// src/routes/myProfile/+page.ts

export const houdini_load = graphql`
	query ViewerInfo { 
		viewer { 
			id
		}
	}
`
```

## Generated type definitions for route functions 

Instead of duplicating the documentation that's already written, you can visit the new typescript guide here: https://docs-git-directory-routes-houdinigraphql.vercel.app/guides/typescript

## Breaking Changes / Notes

- configuration for inline queries (variable functions, hooks, etc.) go in `+page.js`
- inline fragments have a reversed order for arguments 
- `config.sourceGlob` is no longer required and has been deprecated in favor of `config.include` which has a default value that covers most projects 
- added `config.exclude` to filter out files that match the `include` pattern
- `generate --pull-header` is now `generate --header` (abbreviated `-h`)
- `generate  --persist-output` is now `generate --output` (abbreviated `-o`)
- added `schemaPullHeaders` config value to specify the headers sent when pulling the schema

fixes #398, fixes #428, fixes #325, fixes #242, fixes #395, fixes #353, fixes #460
closes #348, closes #452


fehnomenal:
Really nice that you start working on it parallel to their development. Keep up the great work 👍

david-plugge:
Amazing changes, really hyped for the release!

src/runtime/stores/mutation.ts
```diff
@@ -1,43 +1,44 @@
...
+		super()
+		this.artifact = artifact
+		this.store = writable(this.nullState)
+	}
 
-		store.update((c) => {
+	async mutate(
+		variables: _Input,
+		{
+			context,
```
fehnomenal, line 28: I think this `context` can go away. At least it is unused.

fehnomenal (commented):
I'm so eager to try the new houdini that I decided to upgrade sveltekit and houdini (from this branch). A few things I noticed:


### `schemaPath`
I am confused with the `schemaPath`. As it now has a default value I tried removing it. But without `schema` and `schemaPath` I did not get the expected error of "Invalid config file: please provide one of schema or schemaPath. Also, export default config" but only `error when starting dev server: undefined` which is really not helpful. It seems like `throw { filepath, message: '...' }` is incorrect. Well, does not really matter for me as I have two schemas and need `schema*.graphql` to match `schema.graphql` and `schema-system.graphql`.

---

### Client import
Somehow I need to `import { HoudiniClient } from '$houdini/runtime/lib/network';` otherwise `__vite_ssr_import_2__.HoudiniClient is not a constructor TypeError: __vite_ssr_import_2__.HoudiniClient is not a constructor`

---

### Mutations
Defining as in the docs:
```ts
  const login: LoginStore = graphql`
    mutation Login($email: String!, $password: String!) {
      auth_login(email: $email, password: $password, mode: cookie) {
        access_token
        expires
      }
    }
  `;
```

Results in a typescript error:
```
Type 'GraphQLTagResult' is not assignable to type 'LoginStore'.
  Type 'QueryStore<any, any, {}>' is missing the following properties from type 'LoginStore': mutate, nullState ts(2322)
```

But works correctly at runtime.

---

Probably there will be more when I can continue the migration.

AlecAivazis:
Wow @fehnomenal! Really appreciate all this help ❤️  If these are the kinds of comments we're getting we're in a great place already. Was scared the first messages would be something more like "wtf none of this works...." 😅

PimTournaye:
Been eagerly checking out this PR for the past five days, great work!

fehnomenal:
> Wow @fehnomenal! Really appreciate all this help :heart: If these are the kinds of comments we're getting we're in a great place already. Was scared the first messages would be something more like "wtf none of this works...." :sweat_smile:

Well, to be honest at first I struggled a bit :sweat_smile: But after reading the docs I am really happy how smooth everything works :star_struck: Well at least when compared with the sveltekit changes :sweat_smile: 

> Been eagerly checking out this PR for the past five days, great work!

Same, that's why I decided to help getting this released already :joy:

fehnomenal:
I need a way to pass an `authorization` header after logging in. Unfortunately the access token is neither written nor read to/from a cookie by the backend I cannot control (at least not to that extent). So, I have a store that contains the auth session (you guessed it previously it was just `$session`).

I propose to add a generic to `HoudiniClient<RequestData>` -- the name is meh, but there are already multiple contexts. Maybe simply `RequestSession` or `Session`? -- which determines the type of a var that is passed to `fetchQuery`:
```ts
export default new HoudiniClient<{
  accessToken: string | null | undefined;
}>(async ({ requestData, fetch, text, variables }) => {
  const headers: HeadersInit = {
    'content-type': 'application/json',
  };

  if (requestData.accessToken) {
    headers.authorization = `Bearer ${requestData.accessToken}`;
  }

  const response = await fetch('...', {
    method: 'POST',
    headers,
    body: JSON.stringify({ query: text, variables }),
  });

  return await response.json();
});
```

To fill this variable we can use the `.init()` method which still exists but does nothing. It could take a `() => MaybePromise<RequestData>` function and the client would `await` it in `sendRequest` (if present) and pass it to the `fetchFn`.
Then my `+layout.server.ts` and `+layout.ts` could look like this:
```ts
// +layout.server.ts
import { getLoginUrl } from '$lib/utils/auth-urls';
import { redirect } from '@sveltejs/kit';
import type { LayoutServerLoad } from './$types';

export const load: LayoutServerLoad = ({ locals, url }) => {
  // locals.auth is set by `handle`
  if (!locals.auth) {
    throw redirect(302, getLoginUrl(url, undefined));
  }

  return {
    auth: {
      accessToken: locals.auth.accessToken,
      expiresAt: locals.auth.expiresAt.toJSON(),
      isAdmin: locals.auth.isAdmin,
    },
  };
};

// +layout.ts
import houdini from '$lib/houdini';
import type { LayoutLoad } from './$types';

export const load: LayoutLoad = ({ data }) => {
  houdini.init(() => ({
    accessToken: data.auth.accessToken,
  }));
};
```

What do you think? Could this work?

AlecAivazis:
Before I reply to @fehnomenal, I just want to say how crazy it is to see all these reactions and comments. Feels really great to see people are as excited about this as we are. 

Okay, @fehnomenal: you're killing it today. I really like this idea.

It seems like a clean answer to a question that's been looming over me for awhile.  Now that all of the session data is supposed to come from load results, we needed some way to get data out of a load and into a fetch and this does _exactly_ that. It also addresses a lot of the other issues I was struggling with like how do we keep the value up to date, etc.

I think the only real question is if `init` is a good name. Now that "session" is no longer a kit thing, maybe we can take it for ourselves? A quick note about `init` in case it sparks any ideas: atm we need some way of identifying when the initial hydration of your application is finished. There is an [open issue](https://github.com/sveltejs/kit/issues/4447) that is slated to be part of sveltekit's 1.0 but until that's merged we have to use a browser event. This means that every possible route the user might load from needs to have this event listener (even if it doesn't use Houdini). We hide this nasty detail by trying to force users to somehow import the runtime into that first render so we can register the event (this happens in the sveltekit adapter if you're curious). Anyway, the docs say to run `client.init` but all that really matters is that they import `client`, hence why it's now empty an empty function.  My only concern with naming `init` something else is that a user that does not have custom session logic won't be inclined to call `setSession` or whatever and so they won't import the runtime. Having it be called something generic like `init` enforces that it needs to always happen. Maybe we just have 2 different functions for now and then one day init is just no longer "necessary"?

>  there are already multiple contexts

I was thinking about this last night and I'm pretty sure that we can actually merge most of the contexts into one and simplify that part of the codebase considerably. This is something that _constantly_ confuses us when we're talking so it's worth taking the opportunity to simplify it.

---

Both of these are things I want to get in before we actually release `0.16.0` but I dont think they should block the first `@next` version. That being said, I think the `HoudiniClient` import issue is a big problem so I'm going to work on fixing that and then cut a version for more people to test out 🎉

fehnomenal (commented):
### Mutation again

Do we really need the `mutation` function? All it does is wrapping the call to the mutation store with a simpler interface (and obviously the typing will be correct).

But as both calls are really similar, compare
```ts
    const loginStore: LoginStore = graphql`...`;
    const loginFunction = mutation<Login>(graphql`...`);

    const { data } = await loginStore.mutate({ email, password });
    const data = await loginFunction({ email, password });
```
I think it introduces ambiguity which one to use. In the light of the recent routing changes (for which one reason was to reduce ambiguity of how to create routes) I propose nuking the `mutation` function.
Btw the function also hides the nullish type of the `.mutate` call:
image
image
image

---

### Route queries

It seems only `+page.ts` files are handled by the vite plugin but not `+page.server.ts` files. This would be really nice. Are there any technical difficulties against that? Otherwise I can just pass props from `+layout.server.ts` and `+layout.ts` and `await parent()` them.

Also the `...Variable` functions cannot be async but I need it when I have to `await parent()`.

The type of `AfterLoad` is not correctly resolved:
image

For manually loaded queries: How can I access the returned data inside the `load` function? The return value is the normal store. Should I just use `get(Store).data`? It works but `get()` always feels like a hack...

fehnomenal:
Btw I want to make this clear: I absolutely love this project and am so grateful for all your work and effort you pour into your project. I absolutely don't want to discourage you but want to get the best GraphQL client for svelte in existence :smiley:

AlecAivazis:
Oh no problem at all! I **love** the feedback and questions! 

I'm a bit busy right now so my reply might be a little quick but just to address the mutation question: t's mostly there for consistency. 

I had the same question for myself about `query` and `subscription` which both basically just return the store. I figured we'd keep it there for this version while we feel out the new store API and maybe just remove them all together in the next API breaking bump. I'm a bit unsure about what to do with your comments atm. I was going to merge this PR and deploy using the changesets but I don't want to lose your feedback. I guess I'll just deploy from this branch manually once I fix the schema issues tonight 👍

src/cmd/generate.ts
```diff
@@ -43,98 +54,130 @@ export async function runPipeline(config: Config, docs: CollectedGraphQLDocument
...
-	} else if (config.logLevel === LogLevel.Full) {
+		logStyled('CREATED', artifactStats.new, config.logLevel, config.plugin)
+		logStyled('UPDATED', artifactStats.changed, config.logLevel, config.plugin)
+		logStyled('DELETED', artifactStats.deleted, config.logLevel, config.plugin)
+	}
+	// print the status of every file
+	else if (config.logLevel === LogLevel.Full) {
 		for (const artifact of artifactStats.total) {
 			// figure out the emoji to use
 			let emoji = '📃'
```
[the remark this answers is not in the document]
AlecAivazis: 😅 nice catch

AlecAivazis:
Okay, I finally got some time to sit down and go through all this. I'm pretty certain all of the immediately actionable things have been addressed. Thanks again!

> Mutation again

I'm convinced. But I want to take it a step further and also remove `query`, `paginatedQuery`, and `subscription`. Basically, only having the old `fragment` and `paginatedFragment` functions. The three being removed didn't nothing other than return the store and because of that we were being the ambiguity cost you were talking about all over the codebase and documentation. it's better if we just leave only one way of working with graphql documents: `graphql` tags returning Svelte stores.

So that leaves one question: how does a store that's defined inline opt in/out of a generated load? Our current answer is to add a new directive to the pile `@houdini` that we will use moving forward as the place to define one-off configurations (like #239 and #349). For this situation, I suggest we have `@houdini(load: Boolean!)` with a default value of `true`. This means a user can turn the automatic load if they want a lazy inline query. 

> Route queries

There is a technical limitation here. Stores can't be serialized and so they cannot be embedded in the response from a server endpoint. At the moment, I'd like to leave this as a known limitation and we can explore what the right option is. This is an area of kit that's still very much in flux so I'd like to wait a little bit for the dust to settle before we start to jump through a bunch of hoops to hide that limitation from the user.

> The type of AfterLoad is not correctly resolved

You definitely have a version of this release that predates the ✅. The `AfterLoad` type definition has changed in order for the generated `PageData` type to include whatever you add there.  I was honestly shocked to see this work (along with a few other things going on in this release tbh 😅). Anyway, it now looks like this:

```typescript
import { type AfterLoadEvent } from '$houdini'

export function afterLoad({ data }: AfterLoadEvent) {
    return { 
        ...
    }
}
```

AlecAivazis:
I'm going to work on merging this PR tonight so we can release the version with the changesets bot. I started this discussion for future conversations so we can thread messages and what not: https://github.com/HoudiniGraphql/houdini/discussions/475

Really appreciate all of the care and excitement here. It feels like we're really on the break of the something truly special.

fehnomenal:
Re Afterload:

Hmm, I got this generated:
```ts
type AfterLoadReturn = ReturnType<typeof import('./+page').afterLoad>;

type AfterLoadData = {
	GetAvailableWorkshops: GetAvailableWorkshops$result
}

type AfterLoadInput = {
	GetAvailableWorkshops: GetAvailableWorkshops$input
}

export type AfterLoadEvent = {
	event: LoadEvent
	data: AfterLoadData
	input: AfterLoadInput
}

export type AfterLoad = AfterLoadFunction<Params, AfterLoadData, AfterLoadInput>
```
and assumed I can just use `AfterLoad`. This works fine:
```ts
export const afterLoad = ({ data }: AfterLoadEvent) => {
  console.log('afterLoad', data.GetAvailableWorkshops);
};
```
Maybe remove the generated `AfterLoad`? https://github.com/HoudiniGraphql/houdini/blob/dce9397cc93d76dd9adea92f3b8ec5b27dde0bb4/src/cmd/generators/kit/index.ts#L128

Ah well, the `AfterLoadEvent['event']` is still untyped. In the end I will avoid using hooks for all these edge cases. With the new `load_...` functions writing the `load` function manually isn't really that much longer when you need variables and an after hook.

AlecAivazis:
> Ah well, the AfterLoadEvent['event'] is still untyped. 

I'm about to push a fix for this up right now (along with the removal of `query` and the rest). Just cleaning up the documentation

src/runtime/stores/query.ts
```diff
@@ -529,66 +270,275 @@ async function fetchAndCache<_Data extends GraphQLObject, _Input>({
...
+	| LoadEventFetchParams<_Data, _Input>
+	| RequestEventFetchParams<_Data, _Input>
+
+export type QueryResult<_Data, _Input, _Extra = {}> = {
+	data: _Data | null
+	errors: { message: string }[] | null
+	isFetching: boolean
+	partial: boolean
+	source: DataSource | null
+	variables: _Input | null
```
fehnomenal, line 543: This being nullable is strange. When the query has no variables `_Input` is null. If it has variables the `| null` is superfluous.

fehnomenal:
I love all your recent changes. I just noticed one more thing:

Params is `any`
image

I suppose this is because `QueryStoreFetchParams` expects two generic arguments but gets only one:
image

src/cmd/generators/stores/query.ts
```diff
@@ -40,18 +71,22 @@ export default ${storeName}
...
+	constructor() {}
+}
 
-export declare const ${storeName}: QueryStore<${artifactName}$result | undefined, ${VariableInputsType}, ${
-		paginationExtras.storeExtras
-	}> ${paginationExtras.types}
+export const ${globalStoreName}: ${storeName}
 
-export declare const ${config.storeFactoryName(artifactName)}: () => typeof ${storeName}
+export declare const load_${artifactName}: (params: QueryStoreFetchParams<${variableInputsType}>) => Promise<{${artifactName}: ${storeName}}>
```
fehnomenal, line 89:
```suggestion
export declare const load_${artifactName}: (params: QueryStoreFetchParams<${_data}, ${variableInputsType}>) => Promise<{${artifactName}: ${storeName}}>
```

jycouet (dismissed):
Looks great to me 🎉

AlecAivazis:
Okay, i finally got everything building and think this is finally ready to merge. Really appreciate everyone's patience. I think you're really gonna like what we've put together 🎉

PR #473, opened 2022-08-23
Allow setting session data to the client

fehnomenal:
I went ahead and finished my poc for this feature.

Everything is called session which is probably the most accurate name.
Also instead of a function to call the `getSession` is just a prop on the client for simplicity.

### DIFF
--- a/src/runtime/lib/network.ts
+++ b/src/runtime/lib/network.ts
@@ -142,14 +142,23 @@ export type RequestPayload<_Data = any> = {
 /**
  * ## Tip 👇
  *
- * To define types for your metadata, create a file `src/app.d.ts` containing the followingI:
+ * To define types for your metadata, create a file `src/app.d.ts` containing the following:
  *
  * ```ts
- * declare namespace App { *
- * 	interface Metadata {}
+ * declare namespace App {
+ * 	 interface Metadata {}
+ *
+ *   interface SessionData {
+ *     accessToken: string
+ *   }
  * }
  * ```
  *
+ * To define the shape of session data, pass a type or type literal as generic to the client constructor:
+ *
+ * ```ts
+ * export default new HoudiniClient<App.SessionData>(fetchFn);
+ * ```
  */
 export type RequestHandlerArgs<SessionData> = FetchContext &
 	FetchParams & {
--- a/src/runtime/lib/network.ts
+++ b/src/runtime/lib/network.ts
@@ -151,9 +151,10 @@ export type RequestPayload<_Data = any> = {
  * ```
  *
  */
-export type RequestHandlerArgs<SessionData> = Omit<FetchContext & FetchParams, 'stuff'> & {
-	session: SessionData | undefined
-}
+export type RequestHandlerArgs<SessionData> = FetchContext &
+	FetchParams & {
+		session: SessionData | undefined
+	}
 
 export type RequestHandler<_Data, SessionData> = (
 	args: RequestHandlerArgs<SessionData>
--- a/src/runtime/lib/network.ts
+++ b/src/runtime/lib/network.ts
@@ -18,7 +18,7 @@ import {
 export class HoudiniClient<SessionData = undefined> {
 	private fetchFn: RequestHandler<any, SessionData>
 	socket: SubscriptionHandler | null | undefined
-	getSession: (() => SessionData | Promise<SessionData>) | undefined
+	private session: SessionData | undefined
 
 	constructor(
 		networkFn: RequestHandler<any, SessionData>,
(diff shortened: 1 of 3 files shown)


AlecAivazis:
🎉 thanks for putting this together! I think im a bit confused about the API you have in mind since I don't see a way for users to set the session data - am I missing something?

fehnomenal:
Well, it is just an accessor so I thought about:
```ts
// +layout.ts

import houdini from '$lib/houdini';
import type { LayoutLoad } from './$types';

export const load: LayoutLoad = ({ data }) => {
  houdini.init();

  houdini.getSession = () => ({
    accessToken: data.auth.accessToken,
  });

  return {
    isAdmin: data.auth.isAdmin,
  };
};
```

fehnomenal:
Rebased

fehnomenal:
Oh, this would also a good opportunity to update everything mentioning session:
- [ ] authentication guide
- [ ] adapter generation
- [ ] init command

AlecAivazis:
Ahhh okay, I understand now. 

i think making the user assign an async function can lead to kind of an awkward situation if the session data requires an `await`. they'd have to mark that inner function `async` and then there's just a lot of noise left over. What about something like this:

```typescript
// +layout.ts

import houdini from '$lib/houdini';
import type { LayoutLoad } from './$types';

export const load: LayoutLoad = ({ data }) => {
  houdini.init();

  houdini.setSession({
    accessToken: data.auth.accessToken,
  })

  return {
    isAdmin: data.auth.isAdmin,
  };
};
```

that would just set some value in the client that's used during fetch. This has a few side effects:
- the session is only computed once, when it changes (since presumably a change means that load function fires again)
- we avoid accessing any fields on `event` after load is finished. in the past kit has tried to be defensive and throw errors in these situations so it's probably worth grabbing values while we _know_ they're safe.

fehnomenal:
Yes, I agree with you. I will make the changes.

I need to periodically refresh the access token and it will be much easier to do `authSession.subscribe(houdini.setSession)`. Nice idea :+1:

src/runtime/lib/network.ts
```diff
@@ -142,9 +151,13 @@ export type RequestPayload<_Data = any> = {
  * ```
  *
  */
-export type RequestHandlerArgs = Omit<FetchContext & FetchParams, 'stuff'>
+export type RequestHandlerArgs<SessionData> = Omit<FetchContext & FetchParams, 'stuff'> & {
```
AlecAivazis, line 154: we probably want to add a note about `SessionData` in the comment above (can't click on it). Also, while you're here do you mind removing that mention of `stuff`? It's not a field on `FetchParams` anymore

AlecAivazis (changes requested):
👍 really liking this approach.

Another option is having it be a store instead of just a plain object 🤔 It would make it more convenient to render session data but I don't know if that's a good thing. It kind of opens the door to people treating it as just a random place to put things  (although that's already still very possible)

AlecAivazis:
Oh, also don't forget the changeset! No need to add docs if you don't want to, i can figure out where this belongs once its merged

fehnomenal:
I think for now a plain object is enough. Otherwise updating in non-svelte files would be the same level of awkwardness then it is in my example with subscribing.
Time will tell if we need a simpler way of setting the session values.

AlecAivazis:
Oh, we probably want to pass this value to the various load functions, right?

fehnomenal:
You mean the hooks? I hoped to get around touching these code parts :sweat_smile: 

Need to think if this is even necessary...

I guess the `beforeLoad`/`afterLoad` functions should have access to the same scope as the place where you set the session.

<details>
<summary>random brain dump consisting of half sentences and that probably does not make sense but I don't want to forget it if I need it again.</summary>


To correctly use `session` you have to set it on the server (for ssr) and on the client to the same values. So it will probably something returned from one of the load functions (

---

1. *This point is moot as the hooks do not run here*. In an endpoint (querying only on the server): you can do whatever you want and access all values (I think). `session` should nat have changed between requesting and getting the response. Or else it should be trivial to import the same values in the after hook.

---

I can only think of one place where that should not be the case: executions only running on the server. But those cannot be possible because there are no hooks running.

</details>

AlecAivazis:
Ah, crap. It just occured to me that there is a real issue here: this session isn't protected for each user. we either need to find a way to have a per-user store on the server or figure out another approach.... damn.

AlecAivazis:
oh... i didn't know this would close when i merged the other branch - sorry about that!

fehnomenal:
Don't worry this will need a reboot regardless...

PR #476, opened 2022-08-24
Generate input types (and more fixes)

fehnomenal:
See https://github.com/HoudiniGraphql/houdini/discussions/475#discussioncomment-3467838

I also unified and fixed some things I stumbled over.

Does this need a changeset?

### DIFF
--- a/src/cmd/generators/stores/mutation.ts
+++ b/src/cmd/generators/stores/mutation.ts
@@ -29,10 +29,13 @@ export const ${globalStoreName} = new ${storeName}()
 export default ${globalStoreName}
 `
 
-	// type definitions
-	const typeDefs = `import type { ${artifactName}$input, ${artifactName}$result, MutationStore } from '$houdini'
+	const _input = `${artifactName}$input`
+	const _data = `${artifactName}$result`
 
-export declare class ${storeName} extends MutationStore<${artifactName}$result | undefined, ${artifactName}$input>{
+	// the type definitions for the store
+	const typeDefs = `import type { ${_input}, ${_data}, MutationStore } from '$houdini'
+
+export declare class ${storeName} extends MutationStore<${_data} | undefined, ${_input}>{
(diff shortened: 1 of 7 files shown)


AlecAivazis:
No, I dont think this one needs a changeset since the user isn't affected

src/cmd/generators/stores/mutation.ts
```diff
@@ -29,10 +29,13 @@ export const ${globalStoreName} = new ${storeName}()
 export default ${globalStoreName}
 `
 
-	// type definitions
```
AlecAivazis, line 32: I don't think this type is always defined by the typescript generator. Here is the test for a query with no inputs: https://github.com/HoudiniGraphql/houdini/blob/directory-routes/src/cmd/generators/typescript/typescript.test.ts#L233.

Fwiw, I agree it's simpler if we do define the type so I think we just need to tweak the typescript generator and then this should be good to go

AlecAivazis (changes requested):

src/cmd/generators/stores/mutation.ts
fehnomenal: Sorry, I don't get it :sweat_smile: 

The updated test is here and contains the generated input type: https://github.com/fehnomenal/houdini/blob/always-generate-input-types/src/cmd/generators/typescript/typescript.test.ts#L233

The generation code lives in `typescript/index.ts`.

fehnomenal: Ahh, this was just a change to get the prelude to the typescript code in the same shape for queries, mutations, fragments and subscriptions.

AlecAivazis: Ohh you're totally right! Idk how I missed that.

AlecAivazis (approved):
Okay, found some time to pull this down and verify it. Thanks again!

PR #479, opened 2022-08-25
Fix case when readDirSync patch returns object

janvotava:
## Will fix
- readDirSync did not work in the new Svelte Kit because it expects returned `Dirent[]` instead of `string[]`.


jycouet (approved):
Thx for your PR @janvotava 
`LGTM`

janvotava:
> Thx for your PR @janvotava `LGTM`

Thanks for taking care of this PR. Could be probably better - some generic function for handling of arrays of `string[] | Buffer[] | Dirent[]`, but I think for the monkey patch it's kind of all right.

janvotava:
And actually maybe I just instantiate `Dirent` properly through the constructor like:

```
new fs.Dirent('+page.js', fs.constants.UV_DIRENT_FILE)

PR #481, opened 2022-08-25
fix typo: SvelteKit

singingwolfboy:


AlecAivazis (approved):
Thanks for fixing this!

Issue #478, opened 2022-08-25
Incomprehensive Error Message when using `@list` in a query and type not defined


tuxiblo:
### Describe the bug

I had a table and a matching query (external file) and everything worked fine. I deleted the table (in hasura) and forgot to remove the matching query, when running houdini again I got the following error, which I couldn't assign to any file or code changes. 

Query with `@list`:
``` gql
query Organisations {
   Organisation @list(name:"All_Organisations"){
        id
        name
        abbreviation
        legalForm
    }
}
```

Error Message for query with `@list`:
``` bash
> ui@0.0.1 generate
> houdini generate -v -p -ph x-hasura-admin-secret=blablabla

🎩 Generating runtime...
❌ Cannot read properties of undefined (reading 'type')
    at Object.Directive (file:///home/kasimir/Downloads/serverdocumentation/uis/example_ui_one/node_modules/houdini/build/cmd.js:73751:72)
    at Module.visit (/home/kasimir/Downloads/serverdocumentation/uis/example_ui_one/node_modules/graphql/language/visitor.js:243:26)
    at typeCheck (file:///home/kasimir/Downloads/serverdocumentation/uis/example_ui_one/node_modules/houdini/build/cmd.js:73679:17)
    at runPipeline$1 (file:///home/kasimir/Downloads/serverdocumentation/uis/example_ui_one/node_modules/houdini/build/cmd.js:5544:15)
    at runPipeline (file:///home/kasimir/Downloads/serverdocumentation/uis/example_ui_one/node_modules/houdini/build/cmd.js:74421:11)
    at async compile (file:///home/kasimir/Downloads/serverdocumentation/uis/example_ui_one/node_modules/houdini/build/cmd.js:74388:5)
    at async Command.<anonymous> (file:///home/kasimir/Downloads/serverdocumentation/uis/example_ui_one/node_modules/houdini/build/cmd.js:116022:9)

```

After a while, I discovered by coincidence that I have a query that I don't need anymore and deleted it. It then worked again. Further investigation revealed that the error is displayed directly if you do not use `@list ` in your query.

Query wihtout `@list`:
``` gql
query Organisations {
   Organisation {
        id
        name
        abbreviation
        legalForm
    }
}
``` 

Error Message for query without `@list`:
``` bash
> ui@0.0.1 generate
> houdini generate -v -p -ph x-hasura-admin-secret=blablabla

🎩 Generating runtime...
❌ Encountered error in src/graphql/Organisations.graphql
[...]

PR #483, opened 2022-08-25
IMPROVE: Message when @list directive and graphql client type error

jycouet:
## Will fix
- #478


AlecAivazis (changes requested):
nice! we need a changeset for the bot to give us a new version to deploy

AlecAivazis (approved):

PR #489, opened 2022-08-26
Allow passing session data to the fetch function

fehnomenal:
This is the first implementation of #488.

Locally everything works as expected and also client-side refreshes of the session work flawlessly.

## Some highlights
- [ ] New tests are passing
  - [ ]
  - [ ]
- [ ] New settings are available
  - [ ] Name the variable where houdini stores its session. Maybe this is unneeded?
  - [ ]
- [ ] The doc has been updated

--- 

During implementation I noticed that mutations do not take an event object, so I had no easy way to access `event.locals`/`event.parent`. I will tackle this tomorrow.

Probably there should also be more tests.


AlecAivazis:
Closing this in favor of #507 where I'm picking up the torch while you're on vacation. Thanks again for putting all the effort into helping us lately. I hope you're having a relaxing time!

PR #508, opened 2022-09-01
Add sessions

AlecAivazis:
This is a continuation of @fehnomenal's work in #489 that started the support for user-based sessions (ie, authentication). There are 2 parts to this feature: 

1. Session data is set in `hooks.js` using the houdini client's new `setSession(event, { ... })` method
2. Session data is passed to the client's fetch function (similar to how it was previously done)

### What's left
- [ ] Get confirmation from fehnomenal that I didn't miss anything in the transfer
- [x] integration tests for generated layout files (with typecheck for session)
- [x] add warnings (look for `todo`)
- [x] document no more `client.init`
- [x] update init command to not have `client.init`
- [x] update authentication guide
- [x] use `App.Session` type
- [x] `onError` client side navgiation integration tests should fail
- [x] passing session to mutations


endigma:
does this need to use hooks? I think the page data is available to all children? Rich's post on the "removing sessions" discussion mentions doing like:

```svelte
import type { LayoutServerLoad } from "./$types"

export const load: LayoutServerLoad = async function ({ request }) {
	return {
		token: request.headers.get("Authorization"),
	}
}
```

am I missing something?

jycouet:
> does this need to use hooks? I think the page data is available to all children? Rich's post on the "removing sessions" discussion mentions doing like:
> 
> ```svelte
> // src/routes/+layout.server.ts
> 
> import type { LayoutServerLoad } from "./$types"
> 
> export const load: LayoutServerLoad = async function ({ request }) {
> 	return {
> 		token: request.headers.get("Authorization"),
> 	}
> }
> ```
> 
> am I missing something?

You are right, you can get the token like this.

But then, you have to pass it all the time "manually" to do graphql operations.
Like:
```ts
// +page.ts _pseudo code_

export const load: PageLoad = async (event) => {
  const { token } = await event.parent();
  GQL_GetUser.fetch({ variables: xxx, metadata: { token } })
}
```

```ts
// client.ts _pseudo code_

import type { RequestHandlerArgs } from '$houdini'
import { HoudiniClient } from '$houdini'

async function fetchQuery({ fetch, text = '', variables = {}, metadata}: RequestHandlerArgs) {
    const url = 'http://.../graphql'

    const result = await fetch(url, {
        method: 'POST',
        headers: {
            Authorization: `Bearer ${metadata?.token}`
        },
        body: JSON.stringify({
            query: text,
            variables
        })
    })
    return await result.json()
}

export default new HoudiniClient(fetchQuery)
```

You have to do it like this because we can't know that you will name your token. `token`? `superToken`? Maybe you will have a lot of other things in data that are not for Houdini... you will have to `await parent`...

That's a lot of constraints. But it's possible 👍

---

With `hooks` and _`session`_ approach, token authentication is done globally.
And usage becomes lighter focusing on variables & business logic in components.

```ts
// +page.ts _pseudo code_

export const load: PageLoad = async (event) => {
  GQL_GetUser.fetch({ variables: xxx })
}
```

It improves the DX a lot I think 👍
And it's enabling Houdini to generate a lot of boilerplate for you (automatic loading, component queries, ...)

jycouet:
I would like to provide few complete guides later on these common topics 👌
Follow our work 👍

david-plugge:
Nice work!

AlecAivazis:
I'm going to merge this so I can deploy a new version and finally start migrating my applications to the new svelte kit and houdini 🎉  Thanks again everyone for all the help!

endigma:
Docs coming soon?

jycouet:
Actually, all PR has their own docs (check on top "the Vercel preview")

In general `.next` version is available here: https://docs-next-kohl.vercel.app/guides/authentication

PR #514, opened 2022-09-05
Do not force all users to set a session

fehnomenal:
Holy cow, so many exciting updates :joy: Thanks @AlecAivazis for pushing the auth/session stuff to the finish :tada: 

Houdini nagged me to always set a session (otherwise it claims there is an error during `passServerSession`) even though the user is not authenticated. Instead of just `houdini.setSession(undefined)` I decided to clean this up and move the check to the usage of `session` in the fetch function. I found no easy way to test this without creating another testing app that does not set the session but I verified the behavior locally.

I also fixed a bug where the session data was not correctly extracted from `$page.data`. And debatable: I inlined the `passServerSession` method to prevent user confusion.

### DIFF
--- a/src/runtime/lib/network.ts
+++ b/src/runtime/lib/network.ts
@@ -64,13 +64,6 @@ export class HoudiniClient {
 	}
 
 	passServerSession(event: RequestEvent): {} {
-		if (!(sessionKeyName in event.locals)) {
-			// todo: Warn the user that houdini session is not setup correctly.
-			console.log(
-				`Could not find session in event.locals. This should never happen. Please open a ticket on Github and we'll sort it out.`
-			)
-		}
-
 		return {
 			[sessionKeyName]: (event.locals as any)[sessionKeyName],
 		}
--- a/src/runtime/lib/network.ts
+++ b/src/runtime/lib/network.ts
@@ -15,7 +15,7 @@ import {
 	SubscriptionArtifact,
 } from './types'
 
-export const sessionKeyName = 'HOUDINI_SESSION_KEY_NAME'
+export const sessionKeyName = '__houdini__session__'
 
 export class HoudiniClient {
 	private fetchFn: RequestHandler<any>
--- a/src/runtime/lib/network.ts
+++ b/src/runtime/lib/network.ts
@@ -62,12 +62,6 @@ export class HoudiniClient {
 	setSession(event: RequestEvent, session: App.Session) {
 		;(event.locals as any)[sessionKeyName] = session
 	}
-
-	passServerSession(event: RequestEvent): {} {
-		return {
-			[sessionKeyName]: (event.locals as any)[sessionKeyName],
-		}
-	}
 }
 
 export class Environment extends HoudiniClient {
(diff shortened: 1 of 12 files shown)


src/runtime/stores/query.ts
```diff
@@ -6,7 +6,7 @@ import cache from '../cache'
 import type { ConfigFile, QueryArtifact } from '../lib'
 import { deepEquals } from '../lib/deepEquals'
 import * as log from '../lib/log'
-import { fetchQuery, sessionKeyName } from '../lib/network'
+import { fetchQuery, sessionKeyName, sessionSentinel } from '../lib/network'
```
AlecAivazis, line 9: Hm I feel like the `sessionKeyName` and `sessionSentinel` are details that should be hidden away. I know that sessionKeyName was there before but I think we should probably create a few utilities for reading, writing, and checking if the value has been set. What do you think?

src/vite/transforms/kit/init.test.ts
```diff
@@ -15,14 +15,14 @@ test('modifies root +layout.svelte to import adapter', async function () {
...
-		import { setSession } from "$houdini/runtime/lib/network";
+		import { setSession, sessionKeyName } from "$houdini/runtime/lib/network";
 		import { onMount } from "svelte";
 		import { setClientStarted } from "$houdini/runtime/adapter";
 		export let data;
 		onMount(() => setClientStarted());
 
 		page.subscribe(val => {
-		    setSession(val.data);
+		    setSession(val.data[sessionKeyName]);
```
AlecAivazis, line 25: What do you think about moving the `[sessinoKeyName]` into `setSession`? That way the plugin code is a little simpler but also the abstraction isn't quite so leaky.

AlecAivazis (changes requested):
Welcome back! 🎉  Things are getting pretty sweet now, huh? I thought you'd like the removal of `init` 😄  Just a small comment but this should be good to go soon

src/vite/transforms/kit/init.test.ts
fehnomenal: Yeah, I tried that at first, but then the signature of `setSession` would need to look like:

image

and also leaks the key when someone wants to set the client side session manually.

Combined with your other comment this could look like:
```ts
import { extractSession, setSession } from "$houdini/runtime/lib/network";

// ...

setSession(extractSession(val.data));
```

src/runtime/lib/network.ts
```diff
@@ -418,14 +416,45 @@ type KitBeforeLoad = (ctx: BeforeLoadArgs) => Record<string, any> | Promise<Reco
...
+	| App.Session
+> {
+	if (event) {
+		// get the session either from the server side event or the client side event
+		if ('locals' in event) {
+			// this is a server side event (RequestEvent) -> extract the session from locals
+			return extractSession(event.locals as any) || sessionSentinel
+		} else {
+			// this is a client side event -> await the parent data which include the session
+			return extractSession((await event.parent()) as any) || sessionSentinel
```
AlecAivazis, line 455: For a client side event, can't we just look at the page store for data instead of `await`ing the parent?

fehnomenal: Yeah but not inside a `load`. I moved the logic from the `fetchParams` function. The call there already has a check for `isBrowser` and executes without event which means to use the global client side session.

AlecAivazis: Ahhh okay cool i think im following now. Thanks for the quick reply!

PR #517, opened 2022-09-06
Fix unsubscribing inside `load` breaks updates from mutations

fehnomenal:
Fixes https://github.com/HoudiniGraphql/houdini/discussions/475#discussioncomment-3478273

Well this :arrow_up: and the title are ~kind of a stretch because this does not yet fix anything but displays the problem~ now actually the truth :tada:.

What is missing in the comment linked above is that I'm `get($store).data` inside the route `load` function. `get()` from `svelte/stores` subscribes to a store, assigns the data to a variable and immediately unsubscribes. I recreated this logic inside the `load` function here.

In the integration app at `http://127.0.0.1:5173/stores/mutation-update-fragment` a press on the button should increment the counter. But this is only happening after reloading the page.
If you comment the unsubscribe call inside the `load` function everything is working as expected.

I guess the problem lies in `QueryStore.subscribe` or rather the returned unsubscribe handler. Probably this part https://github.com/HoudiniGraphql/houdini/blob/a8b33c11ef91754561f339d289360cdb7126625d/src/runtime/stores/query.ts#L176-L185 But haven't confirmed as I wanted to report this first.

It was actually the problem. Now we only unsub from the cache on the client to respect the comment `// don't clear the store state on the server (breaks SSR)`.

### DIFF
--- a/src/runtime/stores/query.ts
+++ b/src/runtime/stores/query.ts
@@ -175,7 +175,7 @@ If this is leftovers from old versions of houdini, you can safely remove this \`
 			// or when there is still an active subscriber
 			if (this.subscriberCount <= 0) {
 				// clean up any cache subscriptions
-				if (isBrowser && this.subscriptionSpec) {
+				if (clientStarted && isBrowser && this.subscriptionSpec) {
 					cache.unsubscribe(this.subscriptionSpec, this.lastVariables || {})
 				}
 
--- a/.changeset/rotten-peaches-lie.md
+++ b/.changeset/rotten-peaches-lie.md
@@ -1,0 +1,5 @@
+---
+'houdini': patch
+---
+
+Fix unsubscribing inside `load` functions
--- a/integration/api/graphql.mjs
+++ b/integration/api/graphql.mjs
@@ -35,6 +35,8 @@ function getSnapshot(snapshot) {
   return snapshots[snapshot];
 }
 
+let mutationUpdateFragmentDataCounter = 0;
+
 export const resolvers = {
   Query: {
     hello: () => {
@@ -87,7 +89,8 @@ export const resolvers = {
         ...user,
         __typename: 'User'
       };
-    }
+    },
+    mutationUpdateFragmentData: () => ({})
   },
 
   User: {
@@ -132,7 +135,16 @@ export const resolvers = {
         list[userIndex].name = args.name;
       }
       return list[userIndex];
+    },
+    mutationUpdateFragmentInc: () => {
+      mutationUpdateFragmentDataCounter++;
+
+      return {};
     }
+  },
+
+  MutationUpdateFragmentData: {
+    data: () => mutationUpdateFragmentDataCounter
   },
 
   DateTime: new GraphQLScalarType({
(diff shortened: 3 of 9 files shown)


fehnomenal:
Sprinkling in a little `clientStarted` as in 3a04f01 does the trick. hopefully nothing else breaks...

src/runtime/stores/query.ts
```diff
@@ -175,7 +175,7 @@ If this is leftovers from old versions of houdini, you can safely remove this \`
 			// or when there is still an active subscriber
 			if (this.subscriberCount <= 0) {
 				// clean up any cache subscriptions
-				if (isBrowser && this.subscriptionSpec) {
```
AlecAivazis, line 178: It seems a bit off we would let `this.subscriptionSpec` be set but not actually want to unsubscribe. I'm nervous this is a patch over some more subtle issue that's causing us to subscribe to the cache when we shouldn't. We definitely need to subscribe in general, and that first fetch before the client started does need to establish subscriptions so maybe this _is_ the right move?🤔 I'm looking this over lunch so i might be missing something

AlecAivazis: Thinking about this some more, I think the issue is that the first call to `get` creates the subscription when its inside of the load and so the unsubcribe happens before it's actually been sent to the server. I'm not sure I understand why this would be a problem yet but I'll try to find some time to pull this down and investigate.

All of this makes me think that we might want to provide some abstraction over retrieving the store value. That way we can provide it without incurring the subscription cost. My hunch is that we should add a `store.data` that points to the same object that`$store.data` would return. I know it could be confusing have both $store.data and store.data but I think it actually kind of makes sense that the only difference is the `$` since one of them is a reactive reference to the value. What do you think?

fehnomenal: Yeah, I would love this :+1:

fehnomenal: Thinking about this probably it would only make sense for the `load_...` functions right? Elsewhere you would always want to have the current value meaning you need to subscribe.
Following this thought maybe it would be enough to provide a subsubclass of `QueryStore` that exposes a `data` property which is initialized to `await this.fetch(params)` (this already returns the current value of the internal store without touching the cache). I will try this now.

fehnomenal: Alright, this seems to work. I will open a new PR for this.

AlecAivazis: hm good question... it's pretty ugly regardless of how it's implemented so maybe we're going about this wrong. having to always destructure the store name seems kind of convoluted for someone building a manual store. I wonder if it would be better to put something like this in the docs so users feel comfortable using stores directly without relying on load? at least that way getting the value from fetch kind of naturally falls out:

```typescript
    const MySore = new MyQueryStore()
    const { data : value }  = await MyStore.fetch({event})
    
    console.log('do something with', value)

    return { 
        MyStore
    }
  
}
```

AlecAivazis: Having multiple classes that have different capabilities seems like a lot of complexity that i worry would be harder to document than just showing users how to use the fetch result. What do you think?

fehnomenal: Yeah, I ditched the multiple classes approach. And I actually really like the more explicit approach but also the less boilerplate from using `load_...`... Will update the other PR.

PR #520, opened 2022-09-07
Provide the query result for `load_...` functions

fehnomenal:
Alternative to #517 (in the regard of allowing access to the contents of the query store)

This is a really small yet effective change :tada: 

### DIFF
--- a/src/cmd/generators/stores/query.test.ts
+++ b/src/cmd/generators/stores/query.test.ts
@@ -64,7 +64,7 @@ test('basic store', async function () {
 		export async function load_TestQuery(params) {
 			const store = new TestQueryStore()
 
-			await store.fetch(params)
+			store.result = await store.fetch(params)
 
 			return {
 				TestQuery: store,
@@ -112,7 +112,7 @@ test('change globalStorePrefix to "yop___"', async function () {
 		export async function load_TestQuery(params) {
 			const store = new TestQueryStore()
 
-			await store.fetch(params)
+			store.result = await store.fetch(params)
 
 			return {
 				TestQuery: store,
@@ -160,7 +160,7 @@ test('change globalStorePrefix to ""', async function () {
 		export async function load_TestQuery(params) {
 			const store = new TestQueryStore()
 
-			await store.fetch(params)
+			store.result = await store.fetch(params)
 
 			return {
 				TestQuery: store,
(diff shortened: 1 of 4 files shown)


src/cmd/generators/stores/query.test.ts
```diff
@@ -64,7 +64,7 @@ test('basic store', async function () {
 		export async function load_TestQuery(params) {
 			const store = new TestQueryStore()
 
-			await store.fetch(params)
+			store.result = await store.fetch(params)
```
AlecAivazis, line 67: Unforunately this would leave a field that would quickly fall out of date as `fetch` is called in the component. I think a safer approach would be to add a small subscription inside of the query store that updates some local state that is accesible. Something like

```
class QueryStore { 
    data: Data | null = null 

    constructor(...) { 
		this.store = writable(this.initialState)
		this.store.subscribe(val => this.data = val)
    }
}
```

That way you can always use `store.data` and it will always be kept up to date. What do you think?

fehnomenal: Yeah, you're right. The bad thing is that the `result` field is also exposed to the page component...

I thought about many different approaches to keep this scoped to the `load_` functions but in the end a simple way to get the current result without manually subscribing (i.e. using `get_store_value`) won't be possible. At least I think that it would have the same problem that the query unsubscribes from the cache.

AlecAivazis: Hm so we either update `fetchAndCache` to set the local `data` variable that's defined in the previous example or we just make users call `fetch` directly and use the result. I think i prefer the second option even though it's more boilerplate. This is already a very manual situation so I think clarity is the most important thing. The load functions are kind of awkward to use since they are really meant to be mixed into an object. I think that's part of the reason I'm not yet sold on finding a way to make `load_foo` work in this situation

fehnomenal: You know what? I'm more than happy to manually fetch in the few places where I need to check the presence of the fetched data. So, I'm closing this until we find a real use case for this functionality.

AlecAivazis: Okay cool. i'll add some bits to the docs to show people its a blessed pattern. Thanks for thinking through this with me ❤️

PR #534, opened 2022-09-10
More powerful fragment masking

### LINKED ISSUE #349: Add a directive to disable masking

### LINKED ISSUE #529: Dynamic fragment masking
### Describe the feature

Generally I like masking for cleaner separation of concerns and don't want to disable it globally. Right now I have an use case where I would like to disable masking for one specific query:

I use felte for a form to edit a record. The record/form has many fields which can be easily divided into categories (these become individual components and I would like to split the query into fragments corresponding to the components). Felte can take a generic argument to type the form data e.g. for the submit handler. Now just doing `createForm<ThingFromGraphql$result>()` results in the type having only the `$fragments` property and is thus useless for typing...

I would love to be able to write the following:
image
(Probably the directive is at the wrong location but you get the idea)

And instead of `$fragments`
image
there would be all fields in the type.

Would this be possible to implement? I did not yet dive too deep into the artifact generation code. But IIRC this should only affect the generated types, right? As the data returned by the query is complete when logging to the console.

### Criticality

fehnomenal:
fixes #349

I went ahead and implemented a new argument for the `@houdini` directive and was surprised how easy it was.
`@houdini(mask)` is a nullable boolean and can be used on fragment spreads. Depending on the value the generated types for queries, mutations and nested fragments will have the selection set from the fragment or not.

~I added a new config value (`defaultFragmentMasking`) in a backwards compatible way (its value is taken if the `mask` argument is `null` or missing) and deprecated `disableMasking`.~

What is missing is the changeset. This is a new feature so semver wise it would be a minor change bringing us to 0.17. What do you think @AlecAivazis ?


AlecAivazis:
Awesome! I'll try to find some time in the next few days to take a look 👍 

> This is a new feature so semver wise it would be a minor change bringing us to 0.17

before `1.0` we've been using minor versions to denote breaking changes. features and fixes both are being treated as patch versions until we get that official release. Thanks for checking!

site/src/routes/api/config.svx
```diff
@@ -28,7 +28,7 @@ It can contain the following values:
 - `defaultKeys` (optional, default: `["id"]`): A list of fields to use when computing a record's id. The default value is `['id']`. For more information see the [Caching Guide](/guides/caching-data#custom-ids).
 - `types` (optional): an object that customizes the resolution behavior for a specific type. For more information see the [Caching Guide](/guides/caching-data#custom-ids).
 - `logLevel` (optional, default: `"summary"`): Specifies the style of logging houdini will use when generating your file. One of "quiet", "full", "summary", or "short-summary".
-- `disableMasking` (optional, default: `false`): A boolean indicating whether fields from referenced fragments should be included in a document's selection set
```
AlecAivazis, line 31: I agree that the new config value is better but I think we should keep the old one for now just to avoid having a deprecation so soon after breaking so much (fatigue is really real). I think the best bet is to put a note in #459 so we remember to do this as part of the next big release (which it sounds like might be 1.0?? 🤯)

src/cmd/transforms/schema.ts
```diff
@@ -71,9 +71,18 @@ directive @${config.argumentsDirective} on FRAGMENT_DEFINITION
...
+"""
+directive @${config.houdiniDirective}(
+	"""
+	Opt-in to an automatic load function (only valid when used at queries)
+	"""
+	load: Boolean! = true
+	"""
+	Mask fragment fields (only valid when used at a fragment spread)
+	"""
+	mask: Boolean = null
```
AlecAivazis, line 84: I think having `null` kind of confuses things a little bit since it opens the question of how it's different from passing nothing. We should be able to look at the config value to figure out what the appropriate default is.

src/cmd/utils/flattenSelections.ts
```diff
@@ -121,8 +115,34 @@ class FieldCollection {
...
+					case 'NullValue':
+						// use the default from the config
+						break
+
+					case 'BooleanValue':
+						includeFragments = !maskArgument.value.value
+						break
+
+					default:
+						throw new HoudiniError({
```
AlecAivazis, line 137: This should be verified in the typecheck validator. If it isn't, then we want to make sure it is checked there since that validator will report multiple errors at once instead of failing on the first. With that in place (and the nullability removed from my earlier comment), it's okay to do `if (maskArgument?.value.kind === "BooleanValue") ...`

AlecAivazis (changes requested):
Few small tweaks and this should be good to go 👍

src/cmd/transforms/schema.ts
fehnomenal: OK, wow. I was not aware that all arguments in directives seem to be optional even without a default value. Will remove it.

src/cmd/utils/flattenSelections.ts
fehnomenal: Yes, nice :)

site/src/routes/api/config.svx
fehnomenal: The old config is still available and will be used when the new one is not given. It is not an error but will print a warning to the console. Isn't that a nice middleground?

AlecAivazis: Yea, it totally is a good middle ground and If were were at `0.16.5`, I would probably have a different opinion. it just feels not great to have a deprecation warning right after someone has gone through all the trouble to update the other deprecated stuff. And I think `disableMasking` is close enough to reality that it doesn't really justify the potential emotional reaction to having "another thing deprecated"

fehnomenal: Alright, I see you. Are you fine with dedeprecating the old config value or do you want me to remove the new one altogether?

src/cmd/generators/definitions/schema.test.ts
```diff
@@ -204,7 +214,12 @@ test("writing twice doesn't duplicate definitions", async function () {
 		"""@cache is used to specify cache rules for a query"""
 		directive @cache(policy: CachePolicy, partial: Boolean) on QUERY
 
-		"""@houdini is used to configure houdini's internal behavior such as opting-in an automatic load"""
-		directive @houdini(load: Boolean = true) on QUERY
+		"""@houdini is used to configure houdini's internal behavior"""
+		directive @houdini(
+		  """Opt-in to an automatic load function (only valid when used at queries)"""
+		  load: Boolean! = true
```
AlecAivazis, line 220: We will need to remove the `!` on `load` or else it needs to be specified every time `@houdini` is used which is obviously not what we want.

I just tested this with the following node script:

```javascript
import * as graphql from "graphql";

const schema = graphql.buildSchema(`
    directive @Foo(id: ID!) on FIELD

    type Query { 
        test: Int!
    }
`);

console.log(
  graphql.validate(
    schema,
    graphql.parse(`
        { 
            test @Foo
        }
    `)
  )
);

```

fehnomenal: With a default value you can omit it. And that's exactly why I was so confused before your other comment on this file.

I thought the semantics are strange if you specify `@houdini(load: null)`. What will happen? Probably nothing as the check is for `!arg.value.value` but it seems a bit fishy to me :shrug:

AlecAivazis: Okay so what if we do what I originally suggested and have it be non-null with a default pulled from the config value?

```
mask: Boolean! = ${config.disableMasking ? 'false' : 'true'}
```

fehnomenal: Cool idea :+1: I did it but it messed a bit with the formatting...

AlecAivazis (approved):
Thanks again!

PR #539, opened 2022-09-12
:sparkles: NEW: mono repo support

### LINKED ISSUE #536: Houdini 0.16 can't find runtime in monorepo - yarn workspaces
### Describe the bug

Maybe relevant to [this](https://github.com/HoudiniGraphql/houdini/blob/60ecb333a1396f9aa7244eac2f38741a58e7281f/src/common/config.ts#L300)

We have a turbo monorepo structure for our project and attempting to migrate to houdini 0.16 and the new sveltekit, houdini is unable to find the runtime.

```
<root>
node_modules/houdini/xxx
apps/mysveltekitapp/houdini.config.js
```

Our structure is like this, where node_modules are installed in the root but the app lives in a project within the monorepo.

When running `yarn workspace @project/mysveltekitapp houdini generate` it failes with:

```
❌ ENOENT: no such file or directory, stat '/app/apps/mysveltekitapp/node_modules/houdini/build/runtime-esm'
```

It should be looking in the monorepo root, where the runtime exists : `/app/node_modules/houdini/build/runtime-esm`

### Severity

blocking an upgrade

1. Setup a sveltekit project in a monorepo using yarn workspaces
2. Follow houdini getting started guide
3. Error on `houdini generate` in yarn workspace.

### Reproduction

_No response_

jycouet:
Fixes #536

Looking up for `./node_modules/houdini`

- [ ] ~If applicable, add a test that would fail without this fix~
- [ ] ~Make sure the unit and integration tests pass locally with `pnpm run tests` and `cd integration && pnpm run tests`~

### DIFF
--- a/src/common/config.ts
+++ b/src/common/config.ts
@@ -305,7 +305,7 @@ ${
 
 		// if previousLocation !== locationFound that mean that we can go upper
 		// if the directory doesn't exist, let's go upper.
-		while (previousLocation !== locationFound && !fs.isDirectorySync(locationFound)) {
+		while (previousLocation !== locationFound && !fs.existsSync(locationFound)) {
 			// save the previous path
 			previousLocation = locationFound
 
--- a/src/common/config.ts
+++ b/src/common/config.ts
@@ -293,7 +293,7 @@ ${
 		return '$houdini.d.ts'
 	}
 
-	findNodeModulesHoudini(currentLocation: string) {
+	private findNodeModulesHoudini(currentLocation: string) {
 		const pathEndingBy = ['node_modules', 'houdini']
 
 		// Build the first possible location
--- a/src/common/config.ts
+++ b/src/common/config.ts
@@ -293,15 +293,43 @@ ${
 		return '$houdini.d.ts'
 	}
 
+	findNodeModulesHoudini(currentLocation: string) {
+		const pathEndingBy = ['node_modules', 'houdini']
+
+		// Build the first possible location
+		let locationFound = path.join(currentLocation, ...pathEndingBy)
+
+		// previousLocation is nothing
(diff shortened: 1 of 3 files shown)


src/common/fs.ts
```diff
@@ -113,6 +113,10 @@ export async function stat(filepath: string) {
 	return memfs.statSync(filepath)
 }
 
+export function isDirectorySync(dirPath: string) {
+	return fsExtra.existsSync(dirPath)
```
AlecAivazis, line 117: We should add something here to use the mock filesystem in tests

src/common/config.ts
```diff
@@ -293,15 +293,43 @@ ${
 		return '$houdini.d.ts'
 	}
 
+	findNodeModulesHoudini(currentLocation: string) {
```
AlecAivazis, line 296: I think having this as its own public method is a little dangerous since it opens the possibility that someone calls it directly instead of `runtimeSource` which would get around the check for `process.env.TEST`. I think we should either mark this as private or move its logic into `runtimeSource`.

AlecAivazis (changes requested):
🎉

src/common/config.ts
jycouet: Like this?

src/common/fs.ts
jycouet: I added the test, but nit sure about it.
Let me know 👍

AlecAivazis: nice! sorry, one last nitpick - can we call the function `existsSync` since it wraps that function in both scenarios?

jycouet: Yep, of course

AlecAivazis (approved):

PR #542, opened 2022-09-13
Empty client layout load

### LINKED ISSUE #541: Loading in layouts broken?
### Describe the bug

Any query, inline or not, seems to fail when in a `+layout.svelte` or `+layout.ts` file.

### Severity

files:
- `+page.svelte`, content irrelevant
- `+layout.svelte`, inline query of anything and a usage of it
- store will always be undefined

### Reproduction

https://github.com/endigma/houdini-repro

AlecAivazis:
Fixes #541 

After working with @endigma, it turns out that the culprit for this issue was a rather innocent looking load function in a `+layout.ts` file:

```
export async function load() {
    return {}
}
```

I poked around this for awhile but everything I found seemed to indicate that only relying on `+layout.server.ts`  meant that if a load function returns an empty object, we lost the session data. So now, the vite plugin adds a load to the root layout that passes the server's value down to children (who can pick it up from `event.data` before they have to `await parent()`. 

This honestly was a lot of stabbing in the dark since it my expectations were clearly not matching kit's behavior. If @fehnomenal has some time to look at this, I would be super appreciative.


fehnomenal:
Ah yes, I remember reading about that in a kit issue but cannot find it anymore, or was it a discussion :thinking: 

Regardless the gist was that you explicitly have to return `event.data` from `+layout.ts`/`+page.ts` loads.

src/vite/transforms/kit/load.ts
```diff
@@ -2,6 +2,7 @@ import { IdentifierKind } from 'ast-types/gen/kinds'
 import { StatementKind } from 'ast-types/gen/kinds'
 import { namedTypes } from 'ast-types/gen/namedTypes'
 import * as graphql from 'graphql'
+import { config } from 'process'
```
fehnomenal, line 5: Not needed

src/vite/fsPatch.ts
```diff
@@ -21,7 +21,8 @@ export default function HoudiniFsPatch(configFile?: string): Plugin {
 			if (
 				config.isRouteScript(id) ||
 				config.isRootLayout(id) ||
-				config.isRootLayoutServer(id)
+				config.isRootLayoutServer(id) ||
+				id.endsWith('+layout.svelte')
```
fehnomenal, line 25: All changes in this file are probably from debugging, right? Because I don't see why we need to fake every `+layout.svelte`.

Or is it to get rid of many files that only contain `<slot />`?

src/runtime/lib/network.ts
```diff
@@ -448,6 +448,11 @@ export async function getSession(event?: RequestEvent | LoadEvent): Promise<
 		if ('locals' in event) {
 			// this is a server side event (RequestEvent) -> extract the session from locals
 			return extractSession(event.locals as any) || sessionSentinel
+		}
+		// the session data could also already be present in the data field
```
fehnomenal, line 452: When could this be the case? If `getSession` was called inside the root `+layout.ts` load?

src/vite/fsPatch.ts
AlecAivazis: Oh, yea sorry this PR still needs some clean up (i haven't done a self-review yet). Like i said there was a lot of guess work so I still need to trim back what's unnecessary now that I have a decent understanding of the fix.

src/runtime/lib/network.ts
AlecAivazis: This is one of those things that I ran into without a super clear mental model. If I remember correctly, this was to fix sessions in queries found in the root layout but that might not be necessary anymore?

src/vite/fsPatch.ts
fehnomenal: No worries, I fully trust yo to clean up after you :joy:

src/runtime/lib/network.ts
fehnomenal: Well, originally I assumed that `event` is only passed by the user in `load` functions. For `+layout.server.ts` and `+page.server.ts` there is `event.locals` (1st branch). For everywhere else I thought we could just `await parent()` i.e. get it from the root layout.

But yeah, I forgot about queries in the root layout. Here `await parent()` returns only `{}`. So, I think this must stay.

src/vite/fsPatch.ts
AlecAivazis: Okay, so for some reason, if I just have a load that returns an empty function in a `+layout.js` file then I get this error from deep within kit: `node.component is not a function`.  Looking a bit closer, it seems like this is caused by a route that has a script but no view, although it doesn't seem to break a non-houdini project 🤔 

The faked `+layout.svelte` files are there to generate _something_ when there's a layout.js file but it feels like something that really shouldn't be our concern. We're not really touching anything associated with those files in the previous implementation so I'm not sure how we're confusing kit.

AlecAivazis: Its probably worth adding that the vite side doesn't actually result in any "fake" files. it just teaches vite how to resolve them. The actual import and all that stuff comes from kit when it looks at the result of readdirSync

AlecAivazis: Oh, turns out I _do_ see this error in a vanilla kit project so I think it's very much NotOurProblem™️ which feels "right". I'm going to revert these changes and update the integration test.

PR #559, opened 2022-09-20
feat: Config.exclude can be array

### LINKED ISSUE #558: Config.exclude: can be array
### Describe the feature

For more fine grained control, we should be able to pass a string array `exclude` field in `houdini.config.js`.

Maybe `include` also ?

Due to my project structure, I need to pass multiple globs to exclude, like I'm able to do in `graphqlrc.yaml`.

Thank you!

### Criticality

nice to have

DanielHritcu:
# In progress

Fixes #558 

Allow passing an array of strings to `Config.exclude`. Also allow string for this to be a non breaking change.

Should `Config.include` support arrays as well ?

### DIFF
--- a/src/common/config.ts
+++ b/src/common/config.ts
@@ -28,8 +28,8 @@ export class Config {
 	apiUrl?: string
 	schemaPath?: string
 	persistedQueryPath?: string
-	include: string
-	exclude?: string
+	include: string[]
+	exclude: string[]
 	scalars?: ConfigFile['scalars']
 	framework: 'kit' | 'svelte' = 'kit'
 	module: 'commonjs' | 'esm' = 'esm'
@@ -133,8 +133,8 @@ ${
 		this.schemaPath = schemaPath
 		this.apiUrl = apiUrl
 		this.filepath = filepath
-		this.include = include
-		this.exclude = exclude
+		this.include = Array.isArray(include) ? include : [include]
+		this.exclude = exclude ? (Array.isArray(exclude) ? exclude : [exclude]) : []
 		this.framework = framework
 		this.module = module
 		this.projectRoot = path.dirname(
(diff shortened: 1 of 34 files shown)


src/common/config.ts
```diff
@@ -29,7 +29,7 @@ export class Config {
 	schemaPath?: string
 	persistedQueryPath?: string
 	include: string
```
AlecAivazis, line 31: Let's make this same change to include too 👍

AlecAivazis (changes requested):
Thanks for submitting this @DanielHritcu! Mind updating the config docs page too?  It lives in [site/src/routes/api/config.svx](https://github.com/HoudiniGraphql/houdini/blob/main/site/src/routes/api/config.svx).

src/vite/index.ts
```diff
@@ -42,7 +42,10 @@ export default function ({
 					}
 
 					// make sure that the file doesn't match the exclude
-					return !config.exclude || !minimatch(filepath, config.exclude)
+					return (
```
AlecAivazis, line 45: I realize this was there before, but I think we should try to use `config.includeFile` here if possible

src/common/config.ts
DanielHritcu: Will do.

DanielHritcu:
I added the code for include, but I can't manage to build the integration project to test.

I get this weird error, and can't figure out why it's thrown.

The files get included, and the artifacts generated.

Can be replicated starting here: https://github.com/HoudiniGraphql/houdini/pull/559/commits/f149302df95a2b08277a7ea580ca4979ebc4581d

The error I get:
```
file: /Users/danielhritcu/GitHub/houdini/integration/$houdini/runtime/lib/index.js:5:9
3: export { errorsToGraphQLLayout } from './errors';
4: export * from './types';
5: export * as log from './log';
            ^
6: export * from './deepEquals';
7: // putting this here was the only way i could find to reliably avoid import issues
[WebServer] error during build:
Error: Parse error @:5:10
    at parse$b (file:///Users/danielhritcu/GitHub/houdini/node_modules/.pnpm/vite@3.1.0/node_modules/vite/dist/node/chunks/dep-665b0112.js:33612:355)
    at Object.transform (file:///Users/danielhritcu/GitHub/houdini/node_modules/.pnpm/vite@3.1.0/node_modules/vite/dist/node/chunks/dep-665b0112.js:42799:27)
Error: Process from config.webServer was not able to start. Exit code: 1
```

AlecAivazis:
@DanielHritcu are you still running into that error?

DanielHritcu:
@AlecAivazis yes I am.

AlecAivazis:
Oh, interesting it's happening on github too - alright so it _must_ be related to your changes since these tests ran successfully in your other PR. I'll try to figure out what's going on

AlecAivazis:
@DanielHritcu I pushed up a few changes that I think have fixed the problem. I'm really not sure how it's any different from what you have but i'm not getting the errors any more. Let me know what you think

AlecAivazis (approved):

Issue #564, opened 2022-09-22
Local development environment not producable for contribution


sjcobb2022:
### Describe the bug

Hi there,

I'm want to contribute to Houdini however the development environment does not function with the instructions provided on the `[website](https://www.houdinigraphql.com/guides/contributing)`.

If this is of any help: I am on Arch Linux (EndeavorOS).

### Severity

blocking all usage of Houdini

### Steps to Reproduce the Bug

1. Clone the houdini repo
2. Follow instructions on [guides/contributing](https://www.houdinigraphql.com/guides/contributing)
3. `pnpm i && pnpm run build`
4. This gets peer dependency errors (can be ignored with `strict-peer-dependencies=false`). Build also fails.
6. `cd example`
7. `pnpm run generate && pnpm run dev`
8. generate fails and dev fails because of this.

DanielHritcu:
The example fails to build for me also.

Build error:
```
houdini/example/src/routes/[filter]/+page.svelte:4:23
 2 |  import GQL_AddItem from "$houdini/stores/AddItem";
 3 |  import { page } from "$app/stores";
 4 |  import { graphql, type AddItemStore, type AllItemsStore, type NewItemStore } from "$houdini";
                             ^
 5 |  import ItemEntry from "$lib/ItemEntry.svelte";
 6 |  import { derived } from "svelte/store";
error during build:
ParseError: Unexpected token
    at error (file:///houdini/node_modules/.pnpm/svelte@3.49.0/node_modules/svelte/compiler.mjs:17799:19)
```

Dev error:
```
TypeError: __vite_ssr_import_1__.default.init is not a function
    at eval (/src/routes/+layout.js:18:31)
```

I can run the build successfully, but I noticed this in the output:
```
[cmd] [exe] rpt2: options error TS6053: File 'houdini/example/$houdini/graphql/enums.d.ts' not found.
[cmd] [exe]   The file is in the program because:
[cmd] [exe]     Root file specified for compilation
```

DanielHritcu:
@sjcobb2022 
I had an issue with the integration tests. 
But using `nvm` to switch to `node v16.17.0` fixed it.

Just curious, what version of `pnpm` and `node` are you using.
Also, can you post a gist with your error ?

sjcobb2022:
> @sjcobb2022 I had an issue with the integration tests. But using `nvm` to switch to `node v16.17.0` fixed it.
> 
[...]

Issue #549, opened 2022-09-17
schemaPollHeaders non functional


sjcobb2022:
### Describe the bug

Hi there,

I've been spending time trying to load my own headers for an integration with shopify. To access the GraphQL database, you need an API key to be passed as a header. The schemaPollHeaders value is the only way to pass such headers, however it is currently non-functional.

https://github.com/HoudiniGraphql/houdini/blob/e02e524618e518fdf6fcbbb2596a86c9ab9dfe31/src/cmd/pullSchema.ts#L36-L40

The issue of the problem is that the schemaPollHeaders are not combined with the argument headers in the pullSchema function. 

The fix for this could be to combine the objects on line 39 of [src/cmd/pullSchema.ts](https://github.com/HoudiniGraphql/houdini/blob/main/src/cmd/pullSchema.ts#L39)

```js
// Write the schema
await pullSchema(
	config.apiUrl,
	config.schemaPath ? config.schemaPath : path.resolve(targetPath, 'schema.json'),
	{...headers, ...config.pullHeaders}
)
```

Bear in mind, running `houdini generate -h X-Header=TOKEN` works just fine.

### Severity

blocking all usage of Houdini

### Steps to Reproduce the Bug

1. Clone base project with recommended settings (e.g. `npx degit houdinigraphql/intro hello-houdini`)
2. Pass header through schemaPollHeaders in houdini.config.js. example below 
```js
const config = {
	client: './src/client',
	apiUrl: 'https://some.url.com/api/graphql',
	schemaPollHeaders: {
		'X-Necessary-Token-For-Auth': 'env:API_TOKEN'
	}
};
```
3. Unable to access graphql database. Given error: "⚠️  Couldn't pull your latest schema: Unexpected end of JSON input" 

Here is an example config that I believe still works.

```js
const config = {
	client: './src/client',
	apiUrl: 'https://hydrogen-preview.myshopify.com/api/2022-07/graphql.json',
	schemaPollHeaders: {
		'X-Shopify-Storefront-Access-Token': '3b580e70970c4528da70c98e097c2fa0'
	}
};
```

AlecAivazis:
Thanks for submitting this @sjcobb2022! I'll get something out soon. I'm curious, is there a particular reason you are pulling the schema manually instead of letting the vite plugin do it for you?

sjcobb2022:
[...]

PR #566, opened 2022-09-24
Change signature of schema to enable extraConfig when pulling

sjcobb2022:
Addresses possible fix mentioned in #549

Due to not being able to setup the development environment (see #564). I was unable to run pnpm run tests. 

This fix changes the signature of the schema function the vite plugin. This change allows users to define additional user config via the vite plugin. This is beneficial for the end-user as it adds an additional layer of security, through the use of environment variables that are loaded through `.env` files. Currently, must use system-wide environment variables in order to keep data private.

```js
// vite.config.ts
import { sveltekit } from '@sveltejs/kit/vite';
import houdini from 'houdini/vite';
import { defineConfig, loadEnv } from 'vite';

const config = defineConfig(({ mode }) => {
	// loads env files e.g.
	const env = loadEnv(mode, process.cwd());
	return {
		plugins: [houdini({
			schemaPollHeaders: {
				'X-Header': env.API_KEY
			}
		}), sveltekit()]
	};
});

export default config;

```

The above example shows an example config which would allow environment variables to be loaded using vite's loadEnv function. 

Please **do** delete this pull request if you find that my lack of testing is an issue. I would be happy to test this once #564 is resolved, however if anyone has a functional dev environment I would be very happy if they could test it as well.

Sorry and thank you in advance.

### DIFF
--- a/.changeset/mean-mice-help.md
+++ b/.changeset/mean-mice-help.md
@@ -1,0 +1,5 @@
+---
+'houdini': patch
+---
+
+Format Prettier
--- a/.changeset/short-jeans-roll.md
+++ b/.changeset/short-jeans-roll.md
@@ -1,0 +1,5 @@
+---
+'houdini': patch
+---
+
+Change signature of schema function in vite plugin to enable extra config
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: 5.3
+lockfileVersion: 5.4
 
 importers:
     .:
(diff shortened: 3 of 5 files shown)


AlecAivazis:
Thanks for adding this @sjcobb2022 - seems like a totally reasonable thing to do 👍  No need to worry about adding a new test for this, it's very harmless and as long as we didn't break anything we should be good to go

sjcobb2022:
> Thanks for adding this @sjcobb2022 - seems like a totally reasonable thing to do 👍  No need to worry about adding a new test for this, it's very harmless and as long as we didn't break anything we should be good to go

@AlecAivazis should I add a changeset? I can also run prettier.

AlecAivazis:
Thank would be awesome! Thanks @sjcobb2022

sjcobb2022:
Hi @AlecAivazis I accidentally merged main into my fork, however the new changeset is added and formatted with prettier. I can make another pull request later for the integration testing.

sjcobb2022:
@AlecAivazis ok should be good now.

.changeset/mean-mice-help.md
```diff
@@ -0,0 +1,5 @@
+---
+'houdini': patch
```
AlecAivazis, line 2: For future reference, you only need a changeset if the update affects the API in some way. Actions like running prettier don't need to be part of the changelog since they are totally internal.

No need to change anything, I can take it from here but I thought I'd let you know for the future.

AlecAivazis (approved):

.changeset/mean-mice-help.md
sjcobb2022: Oops sorry. Thanks for the info!

PR #576, opened 2022-09-28
Nested list cache bug

wattsjs:
Hey @Alec

I'm not sure what the data format is for these queries but is there some nested information available on the User that we can use to test this? In this example **only** the nested->something value is mutated on the parent store when the child query changes, and the other values (e.g. id, name) are fine


AlecAivazis:
Oh, I see now 👍  yea there is a `friendsList` that points to a list of users. If we need to add a non-list version we can do that too

AlecAivazis:
I'm going to merge this in and try to take it from here - thanks again!

PR #584, opened 2022-10-03
Added further config to vite plugin

sjcobb2022:
Addition to #566

After testing, there are some instances where the config is not passed into other functions. By adding extraconfig to all functions in vite, we are able to ensure that any additional config is always passed.

Sorry for all the little changes @AlecAivazis

### DIFF
--- a/.changeset/two-papayas-beg.md
+++ b/.changeset/two-papayas-beg.md
@@ -1,0 +1,5 @@
+---
+"houdini": patch
+---
+
+Added further config to vite plugin


AlecAivazis:
> Sorry for all the little changes @AlecAivazis 

No worries at all! I'd rather the little changes sprinkle in than not get caught :D  Thanks for taking the time to fix it 👍

AlecAivazis:
Mind adding a changeset?

AlecAivazis (approved):

AlecAivazis:
Thanks again!

PR #590, opened 2022-10-04
added extra config info to docs

sjcobb2022:
Docs edit referenced in #589

Added information about using .env files during the vite preprocessing step.

Not sure if I need to add a changeset so I haven't yet. Please inform me if I do, as it doesn't directly affect the user API.

I also just placed this at the bottom. Please feel free to edit  the order to your liking.

### DIFF
--- a/site/src/routes/guides/faq.svx
+++ b/site/src/routes/guides/faq.svx
@@ -94,17 +94,6 @@ If you have a route directory that does not contain a `+page.js` file and does n
 plugin will generate an empty `+page.js` file that is never used. This confuses vite hence the warning. Don't worry - since this empty
 file is never imported there are no performance implications.
 
-### My IDE is complaining that the internal directives and fragments don't exist.
-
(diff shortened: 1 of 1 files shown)


AlecAivazis:
Looks great! No need for a changeset 👍 

I think we can probably put it higher on the list. How about right before "What is this Generated an ....."

sjcobb2022:
> Looks great! No need for a changeset +1
> 
> I think we can probably put it higher on the list. How about right before `What is this Generated an .....`

I can do that now!

sjcobb2022:
Alright all formatted correctly now @AlecAivazis

site/src/routes/guides/faq.svx
```diff
@@ -88,6 +88,37 @@ If you haven't seen Houdini's document stores before, please check out the [Work
 You should use the `metadata` parameter in the document store to pass arbitrary information into your client's network function. For more information,
 please visit [the query store docs](/api/query#passing-metadata).
 
+### Can I use .env files for pre-processing?
+
+Yes you can! The vite plugin supports passing additional config during the pre-processing stage. Any config present in the [houdini config](/api/config#fields) is a valid input.
```
AlecAivazis, line 93:
```suggestion
Yes you can! You can pass any [valid configuration value](/api/config#fields) to the vite plugin to use your environment variable:
```

AlecAivazis (changes requested):
Thanks for moving that. Quick tweak on the wording and we can merge this

AlecAivazis (approved):

PR #601, opened 2022-10-11
Svelte layout gql

jycouet:
Fixes #TICKET


packages/houdini-svelte/src/plugin/transforms/kit/load.ts
```diff
@@ -423,7 +431,7 @@ async function find_page_query(page: SvelteTransformPage): Promise<LoadTarget |
 	) as graphql.OperationDefinitionNode
 	// if it doesn't exist, there is an error, but no discovered query either
 	if (!definition) {
-		formatErrors({ message: 'page.gql must contain a query.', filepath: page_query_path })
+		formatErrors({ message: 'gql file must contain a query.', filepath: page_query_path })
```
AlecAivazis, line 434: This error message is a little confusing. Can we use the `type` or `page` args to figure out if we are analyzing a page or layout file? Also, we probably want to use the pageQueryFileName config here (and add one for layout queries?)

jycouet: Now, `filepath: query_path` so people will see the file that has an issue.
I guess it's sufficient?

AlecAivazis (approved):

PR #627, opened 2022-10-21
doc update after my migration

jycouet:
After my migration, I propose a few tweak doc update to make it even clearer

### DIFF
--- a/site/src/routes/guides/release-notes.svx
+++ b/site/src/routes/guides/release-notes.svx
@@ -24,24 +24,26 @@ The example below shows every config value that needs to be moved:
 
   /** @type {import('houdini').ConfigFile} */
   const config = {
+		apiUrl: 'http://localhost:4000/graphql'
+
 -   client: './src/client.ts'
--   pageQueryFilename: '+myQuery.gql'
--   globalStorePrefix: "G_",
--   quietQueryErrors: true,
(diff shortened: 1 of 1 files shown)


site/src/routes/guides/release-notes.svx
```diff
@@ -24,41 +24,43 @@ The example below shows every config value that needs to be moved:
 
   /** @type {import('houdini').ConfigFile} */
   const config = {
+    apiUrl: 'http://localhost:4000/graphql'
```
AlecAivazis, line 27: I'm not sure it's worth complicating these examples with config that doesn't matter. I think we should keep it as it is since I think it could confuse people. Thanks for opening the PR tho!

AlecAivazis:
Whoops! Just realized i closed this by mistake 😅

site/src/routes/guides/release-notes.svx
jycouet: Actually, I think that it's not complicating it, but simplifying it.

- Showing 1 prop not moving on top as an example
- Showing less props in `plugins.houdini-svelte` example
- Showing that scalar didn't move (and making it explicite)
- Having the link for all config at the end

AlecAivazis (approved):

PR #633, opened 2022-10-22
Fix/double query

### LINKED ISSUE #632: +page.gql are sent twice
### Discussed in https://github.com/HoudiniGraphql/houdini/discussions/609

<sup>Originally posted by **philippxbt** October 17, 2022</sup>
Hi, 
I'm using automatic loading for my queries by adding a +page.gql file in the respective route directories. I have also defined the variables functions, because all my queries have variables. 

The queries are working, however, I noticed that each graphql request is sent twice to the server, the first request without variables and the second one with the variables. 

Is this default behavior or am I doing something wrong?</div>

A simple way to reproduce this is to have:
```gql

# +page.gql

query ProjectsList($limit: Int) {
  project(limit: $limit) {
    username
}
```
And add a log in your client `fetchQuery`. You will see that it will be fired 2 times.

This is because the preprocessor will generate code in: `+page.js` and `+layout.js`.

To fix this:
- `+page.gql` should only affect `+page.js` 
- `+layout.gql` should only affect `+layout.js`

jycouet:
Fixes #632

The preprocessor generates code in: +page.js and +layout.js.
This PR is to fix this
- +page.gql generates only +page.js
- +layout.gql generates only +layout.js


packages/houdini-svelte/src/plugin/transforms/kit/load.test.ts
```diff
@@ -477,6 +478,32 @@ describe('kit route processor', function () {
 			    };
 			}
 		`)
+
+		expect(route.layout).toMatchInlineSnapshot(`
```
AlecAivazis, line 482: This seems a bit strange but maybe it's just a side effect of the filepath it was mocked at. Just a reminder for me to investigate

packages/houdini-svelte/src/plugin/kit.ts
```diff
@@ -76,6 +76,16 @@ export function is_layout_component(config: Config, filename: string) {
 	)
 }
 
+export function is_layout_something(config: Config, filename: string) {
+	return (
+		is_layout_script('kit', filename) ||
```
AlecAivazis, line 81: I think we probably want to avoid hardcoding `kit` here. the rest of the utils like `is_component` ask for the `Framework` as a saftey measure so we should probably do the same here

AlecAivazis (changes requested):

PR #641, opened 2022-10-24
:bug: FIX: tsconfig not written in init cmd due to comment in new kit…

jycouet:
… version

Fixes #TICKET

### DIFF
--- a/packages/houdini/src/cmd/init.ts
+++ b/packages/houdini/src/cmd/init.ts
@@ -2,7 +2,7 @@ import { getIntrospectionQuery } from 'graphql'
 import fetch from 'node-fetch'
 import prompts from 'prompts'
 
-import { fs, path, pullSchema } from '../lib'
+import { fs, parseJSON, path, pullSchema } from '../lib'
 import { ConfigFile } from '../runtime/lib/config'
 
 // the init command is responsible for scaffolding a few files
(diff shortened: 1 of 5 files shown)


AlecAivazis:
I'm starting to think we're being too smart in init. What do you think if we just always write the new file instead of trying to figure out if it's "safe". If we destroy users code they set up, they can always look at the diff to get it back... Thoughts?

packages/houdini/src/cmd/init.ts
```diff
@@ -291,8 +291,13 @@ async function tjsConfig(targetPath: string, framework: 'kit' | 'svelte') {
 
 	// check if the tsconfig.json file exists
 	try {
-		const tjsConfigFile = await fs.readFile(configFile)
+		let tjsConfigFile = await fs.readFile(configFile)
 		if (tjsConfigFile) {
+			// remove all comments to be able to parse the file, and add stuff to it.
+			tjsConfigFile = tjsConfigFile.replace(
+				/\\"|"(?:\\"|[^"])*"|(\/\/.*|\/\*[\s\S]*?\*\/)/g,
```
AlecAivazis, line 298: I feel like we should have a test that verifies this regex

jycouet:
I'm totally for it. 

1/ Check if nothing is pending in GIT
2/ Questions
3/ Write files

Mermaid 👇

```mermaid
flowchart LR

A[init] -->B(GIT Status)
B --> |not clean| A
B --> C{is Running}
C --> |no| D{schemaPath}
D --> |not correct| D
D --> F(done)
C --> |yes| E{url}
E --> |not correct| E
E --> F(DONE)
```

---

I think that this PR could be merged to fix today's issue.

And tackle the "bigger view" in a separate PR with also plugin ideas?

AlecAivazis (approved):
Yea that sounds good! Thanks for adding the test 👍

AlecAivazis:
Yea I like the additional branch for a local schema path

Issue #653, opened 2022-10-26
PageData is not a single type


samuel-utbult-oborgen:
### Describe the bug

Suppose your app.d.ts file looks like this:

```
declare namespace App {
	// interface Locals {}
	interface PageData {
		valueFromLoad: string;
	}
	// interface Error {}
	// interface Platform {}
}
```

And suppose you have a layout file like this:

```
export function load(event: LoadEvent) {
    return {
        valueFromLoad: "Value from load"
    };
}
```

Then, you would have a view looking like this:

```
<script lang="ts">
    import type { PageData } from "./$houdini";

    export let data: PageData;

    $: ({ GetEntries, valueFromLoad } = data);
</script>

{$valueFromLoad}

// And use $GetEntries.data somehow.
```

This is possible to run but TypeScript complains that valueFromLoad cannot be found even though it has been specified in App.PageData. I can circumvent this problem by writing this:

```
<script lang="ts">
    import type { PageData as HoudiniPageData } from "./$houdini";
    import type { PageData } from "./$types";

    export let data: any;

    $: ({ GetShips } = data as HoudiniPageData);
    $: ({ valueFromLoad } = data as PageData);
</script>
```

When inspecting the corresponding `$houdini.d.ts` file, I see this:

```
import type { PageLoadEvent, PageData as KitPageData } from './$types'

export type PageData = {
		GetShips: GetShipsStore
}
```

`KitPageData` is what I referred to above as `PageData`. Shouldn't the file look something like the code below?

```
import type { PageLoadEvent, PageData as KitPageData } from './$types'

export type PageData = KitPageData & {
		GetShips: GetShipsStore
}
```

I wanted to give you a Stackbliz link but I can't get any public API I found to work. I hope this description makes it possible to reproduce the problem though.

### Severity

serious, but I can work around it

### Steps to Reproduce the Bug

See above.

jycouet:
Hello,
Thx for reaching out.

In general (not houdini specific) I think that it's not a good idea to define `PageData` in `app.d.ts` because you will probably have more than one route with more than one kind of type to return.
In `kit`, you will get automatically `PageData` within each route.

If you are interested in having your data + houdini's store, you could have a look at https://www.houdinigraphql.com/api/query "Manual Loading".

Let us know if I missed something. 👍


_Side note, I use the star wars public api to demo things like in https://github.com/jycouet/svienna_22_10_

AlecAivazis:
[...]

PR #673, opened 2022-11-01
Route typegen changes - Sveltekit Page/LayoutData mixin + Layout houdini_load Inheritance

sjcobb2022:
Fixes #608 #653

Hi there, here is the initial pull for the new type generation that I've been working on. It currently works for all of the cases that are provided in the e2e tests. I haven't had the time to modify any of the type generation tests yet, so I don't think this will pass CI.

Instead of generating a new file, our types now mirror svelte-kit's types which enables both the inheritance of `houdini_load`s in layout groups, and automatically inherits the data of SkelteKits load functions. 

This should probably be a draft PR as there are probably some code-formatting preferences that need to be looked at, I would appreciate any changes that are made.

P.S. not sure if I updated all the right packages correctly in the changeset. Please feel free to change/fix it :)).


sjcobb2022:
Hi @AlecAivazis.

Not sure why the test failed for windows in the playwright tests. I will make another commit in to ensure that it wasn't a failed merge or something. I'll format as well in a bit.

There are some issues with the current generation since, we it's dependant on the .svelte-kit folder. Since this is this is the case, we either need to call npx svelte-kit sync in our mock dir or seperate the type tests out.

sjcobb2022:
Ok @AlecAivazis @jycouet, All the tests now pass and the e2e on windows is no longer being funny. I think its all good to go.

AlecAivazis:
Thanks for putting this together @sjcobb2022! I really appreciate the effort. It might take me a few days to dig into this in more detail since it's such a large block of code. Hopefully i can find some time before the weekend

sjcobb2022:
> Thanks for putting this together @sjcobb2022! I really appreciate the effort. It might take me a few days to dig into this in more detail since it's such a large block of code. Hopefully i can find some time before the weekend

Most of it is the package-lock.json which I think got messed up by a package being moved (I think it was when i ran npx playwright install). It's "only" a diff of about 800 lines if this is the case.

packages/houdini/src/lib/fs.ts
```diff
@@ -6,6 +6,19 @@ import { promisify } from 'util'
 
 import * as path from './path'
 
+export async function copySync(srcDir: string, destDir: string, overwrite: boolean = false) {
```
AlecAivazis, line 9: all of the functions in this module need to be safe to use in unit tests. You can check for `process.env.NODE_ENV` to see if you need to copy to the memory fs or the real one

packages/houdini-svelte/src/plugin/codegen/routes/index.ts
```diff
@@ -20,170 +32,318 @@ export default async function svelteKitGenerator(
...
-			// pageScript need to be imported so we can figure out if there is a houdini_load
-			// and what's inside.
-			if (pageScript) {
-				// import the houdini_load function
-				const { houdini_load, exports } = await extract_load_function(config, pageScript)
-
-				// add every load to the list
-				pageQueries.push(...(houdini_load ?? []))
-				scriptPageExports = exports
+	async function walk_types(dirpath: string) {
```
AlecAivazis, line 35: The body of this function looks very simliar to `walk_routes`, am i missing something?

packages/houdini-svelte/src/plugin/codegen/routes/index.ts
```diff
@@ -20,170 +32,318 @@ export default async function svelteKitGenerator(
...
+			const skTypeFile = path.join(
+				config.projectRoot,
+				'.svelte-kit/types',
+				dirpath.match(relative_path_regex)?.[0] ?? '',
+				'$types.d.ts'
 			)
 
-			// make sure we have a home for the directory
-			await fs.mkdirp(path.dirname(target))
+			if (fs.existsSync(skTypeFile)) {
```
AlecAivazis, line 169: I think this block of code could use some comments.  its kind of tricky to follow what's goin going on without understanding the general approach

AlecAivazis (commented):
Made a very shallow pass at this. Will find some time to look in more detail (which the extra comments will help with)

packages/houdini-svelte/src/plugin/codegen/routes/index.ts
sjcobb2022: I found that walk routes felt overly complex and was trying to deal with too much for this case. Since we are merely copying the svelte types, we have a lot more information anyway, meaning that most extra inputs are unnecessary. 

Structurally, I think it would be good to move this into kit.ts.

packages/houdini/src/lib/fs.ts
sjcobb2022: Ok I'll update this.

sjcobb2022: > all of the functions in this module need to be safe to use in unit tests. You can check for `process.env.NODE_ENV` to see if you need to copy to the memory fs or the real one

Is it alright if I add memfs-extra as a dep then? memfs by itslef doesn't ahve copy sync.

sjcobb2022: @AlecAivazis I just realised that the actual code does not implement either of the exports. I think I added these during development and then forgot about them.

sjcobb2022:
@AlecAivazis I've resolved the comments and other things. What direction would you go for the use of walk_routes/walk_types. I'm unsure of what to do.

AlecAivazis (commented):
Thanks for adding comments and cleaning up the exports!

sjcobb2022:
> Thanks for adding comments and cleaning up the exports!

I think I understand what you mean. I think that if we were to use walk_routes we could do a very similar process.

If we modified walk_routes to the have the body of walk_types like so:

https://github.com/HoudiniGraphql/houdini/blob/b1f276d729cc994886eae1b83ba33d378da92b7c/packages/houdini-svelte/src/plugin/codegen/routes/index.ts#L36-L143

You would then have to change the definition of our route function so that it would take in a few things.

The routes function would then need the pageQueries, layoutQueries, pageExports, and layoutExports, and the dirpath.

sjcobb2022:
@AlecAivazis I implemented the major changes cause I had some free time. Hope this helps.

sjcobb2022:
Sorry @AlecAivazis there was a merge issue that cause pnpm to get out of sync. All OK now.

jycouet:
@sjcobb2022 thx a lot for your work, I had a look, and it will be a great addition.

Looking at the route `src/routes/stores/ssr/*`, I think there is an issue of import path (check the screenshot):
image

Could you check it? I could also have a look if you want to.

Thx again for the effort 👍

sjcobb2022:
> @sjcobb2022 thx a lot for your work, I had a look, and it will be a great addition.
> 
> Looking at the route `src/routes/stores/ssr/*`, I think there is an issue of import path (check the screenshot):
> image
> 
> Could you check it? I could also have a look if you want to.
> 
> Thx again for the effort 👍

Ah yes you're right. I think we need to copy the proxy.js files that sveltekit produces.

sjcobb2022:
Awesome. @jycouet pretty easy fix, just needed to check for the proxy file and copy if not. I changes the fs file but I maintained the same format as the other functions (ensuring memory safety during tests or whatever the idea was). I hope this is alright.

jycouet (approved):
Thank you @sjcobb2022 for your efficiency, you fixed it really quick 👍
On my side, after testing the feature everything looks good.

AlecAivazis:
Really appreciate you verifying this @jycouet ❤️ I'll give it a final once over tomorrow and get this merged finally!

sjcobb2022:
@AlecAivazis Sorry I accidentally closed the pull. Also the windows did it's 6000ms polling thing and needs to be rerun I think.

AlecAivazis:
Heh i actually restarted the windows one when I saw it didn't pass 😅  I'll definitely have some time to give this a final pass today

sjcobb2022:
@AlecAivazis Thanks man.

AlecAivazis (commented):
Just a few last minute nitpicks, tried to cover as much as I could with suggestions but i couldn't get them all

sjcobb2022:
@AlecAivazis I think this is everything?

AlecAivazis (approved):
Awesome!! Thanks for doing that 👍

PR #681, opened 2022-11-07
Provide parentType to addMany subscriptions

### LINKED ISSUE #360: Add support for multiple list directives in the same document
Currently you can only use one `@paginate` in a single document and if you have `@paginate` and `@list` in the same document you can run into trouble when the pagination methods trigger.

Joklost:
Partial fix for #360

The parentType in addFieldSubscription was hardcoded to 'asdf'. This causes issues when using multiple nested @list directives, as using the [ListName]_(insert|remove|toggle) operations with @parentID attempted to look for a parent with type 'asdf'. 


Joklost:
Darn it, formatting fails :P

Just pushed the changes - will open another one..

AlecAivazis:
_sigh_ so sorry about all this. I really have no idea what's going on with GitHub....

PR #682, opened 2022-11-07
Provide parentType to addMany subscriptions

Joklost:
Partial fix for #360

The parentType in addFieldSubscription was hardcoded to 'asdf'. This causes issues when using multiple nested @list directives, as using the [ListName]_(insert|remove|toggle) operations with @parentID attempted to look for a parent with type 'asdf'. 

Hopefully last attempt!


packages/houdini/src/runtime/cache/subscription.ts
```diff
@@ -175,14 +176,15 @@ export class InMemorySubscriptions {
 					key,
 					selection: fieldSelection,
 					spec,
-					parentType: 'asdf',
```
AlecAivazis, line 178: So the tests still seem to pass if this is reverted to the incorrect value. Ill try to find some time today to update the test and merge this in.

Joklost: Hmm, interesting.. Thanks!

AlecAivazis: Well finding the edge case was non-trivial. Anyway, thanks again for fixing this!

AlecAivazis (approved):

AlecAivazis (approved):

packages/houdini/src/runtime/cache/subscription.ts
Joklost: No problem - and thanks for assisting with the test case!

PR #693, opened 2022-11-15
Feat/lists all

jycouet:
Fixes #TICKET


packages/houdini/src/codegen/generators/artifacts/operations.ts
```diff
@@ -113,20 +114,42 @@ function operationObject({
 		const append = internalDirectives.find(
 			({ name }) => name.value === config.listAppendDirective
 		)
-		// is when applied?
-		const when = internalDirectives.find(({ name }) => name.value === 'when')
-		// is when_not applied?
-		const when_not = internalDirectives.find(({ name }) => name.value === 'when_not')
+		// if both are applied, there's a problem
+		if (append && prepend) {
```
AlecAivazis, line 118: These errors shoud be caught in a validator if possible so that we can report multiple of them at the same time instead of failing on the first. Same for using `@parentID` and `@allLists` together

AlecAivazis (changes requested):

packages/houdini/src/codegen/generators/artifacts/artifacts.test.ts
```diff
@@ -869,6 +869,293 @@ describe('mutation artifacts', function () {
...
+		const queryContents = await fs.readFile(
+			path.join(config.artifactPath(mutationDocs[0].document))
+		)
+		expect(queryContents).toBeTruthy()
+		// parse the contents
+		const parsedQuery: ProgramKind = recast.parse(queryContents!, {
+			parser: typeScriptParser,
+		}).program
+		// verify contents
+		expect(parsedQuery).toMatchInlineSnapshot(`
```
AlecAivazis, line 905: This test can be simplified  using the custom matcher. The test you copied from was an old one 😅 

```suggestion

		expect(mutationDocs[0]).toMatchInlineSnapshot(`
```

packages/houdini/src/runtime/lib/config.ts
```diff
@@ -125,6 +125,16 @@ export type ConfigFile = {
...
 
+	/**
+	 * Specifies whether mutations should append or prepend list. For more information: https://www.houdinigraphql.com/api/graphql (default: `append`)
+	 */
+	defaultListPosition?: 'append' | 'prepend'
+
+	/**
+	 * Specifies whether mutation should apply to all list (default: `false`)
+	 */
+	defaultListAllList?: boolean
```
AlecAivazis, line 136: This config name is a bit strange to me. Not quite sure what a better alternative is 🤔

AlecAivazis: maybe `config.defaultListTarget: 'all' | null`

packages/houdini/src/codegen/generators/artifacts/operations.ts
jycouet: I moved it to the `typeChecks` and added tests to it.

AlecAivazis (approved):
Looks good! Thanks again ❤️

PR #708, opened 2022-11-18
FIX: missing VariableFunction type

jycouet:
Improve the detection of `VariableFunction` needed.

- [x] ~Update the first line to point to the ticket that this PR fixes~
- [x] ~If applicable, add a test that would fail without this fix~

### DIFF
--- a/.changeset/flat-donuts-itch.md
+++ b/.changeset/flat-donuts-itch.md
@@ -1,0 +1,5 @@
+---
+'houdini-svelte': patch
+---
+
+import VariableFunction type when a function type is defined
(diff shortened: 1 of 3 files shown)


jycouet:
@sjcobb2022 you might want to have a look also?

AlecAivazis:
This seems to work for me. I'll wait a bit for @sjcobb2022's 👍 for a bit

sjcobb2022:
> @sjcobb2022 you might want to have a look also?

Looks good to me! This covers more edge cases so definitely a bonus.

AlecAivazis (approved):

PR #717, opened 2022-11-19
Query stores & initial state.

jycouet:
In a fake await scenario, let's return the cached value in the store (if policy & co is ok with that) first, then do the normal flow.


packages/houdini-svelte/src/plugin/codegen/stores/query.ts
```diff
@@ -76,16 +74,62 @@ export default ${globalStoreName}
 	const typeDefs = `import type { ${_input}, ${_data}, ${store_class}, QueryStoreFetchParams} from '$houdini'
 
 export declare class ${storeName} extends ${store_class}<${_data}, ${_input}> {
+	/**
+	 * ### Default usage
```
AlecAivazis, line 78: We probably want to add something here for manual loads that want to use the value of `fetch` since that's one of the documented uses of the class constructor

packages/houdini-svelte/src/runtime/stores/query.ts
```diff
@@ -155,6 +157,21 @@ If this is leftovers from old versions of houdini, you can safely remove this \`
 		// if the await isn't fake, await it
 		if (!fakeAwait) {
```
AlecAivazis, line 158: Is there a reason not to set the cache before the request even if its not a fake await?

packages/houdini/src/runtime/lib/network.ts
```diff
@@ -234,8 +234,52 @@ export async function fetchQuery<_Data extends GraphQLObject, _Input extends {}>
...
+	})
+
+	return {
+		result: result.body,
+		source: result.ssr ? DataSource.Ssr : DataSource.Network,
+		partial: false,
+	}
+}
+
+export async function fetchCache<_Data extends GraphQLObject, _Input extends {}>({
```
AlecAivazis, line 272: The name of this function is pretty easily confused with `fetchAndCache`. Between, `fetchQuery`, `fetchAndCache` and `fetchCache` it'd be tough for someone to know what the right one to use is

packages/houdini-svelte/src/plugin/codegen/stores/query.ts
jycouet: Yes sure, what do you suggest?

packages/houdini-svelte/src/runtime/stores/query.ts
jycouet: The idea was that if `await` is allowed, we have the time, and the same check will be done in `fetchAndCache`.
But, true, I don't like to see it after the request.

I will move it up 👍

packages/houdini/src/runtime/lib/network.ts
jycouet: Yeah, agree that it can be confusing. 
`getFromCache`? `getCache`? `retreiveCache`? other suggestions?

AlecAivazis (approved):

PR #719, opened 2022-11-19
FIX: page/load BeforeLoadEvent types

### LINKED ISSUE #718: onError and beforeLoad cause issues due to relative imports.
### Describe the bug

Currently there is a minor bug that adds the type `[string: any]` to the `PageData` and `LayoutData` types if a beforeLoad is in a layout route. Didn't catch this due to minor mistake in test writing.

This will cause an error since if a BeforeLoadEvent for the layout references the PageLoadEvent, there is a circular dependency that typescript can't deal with. 

There is a similar case for AfterLoadEvent type however this isn't as serious.

I have a fix in the works but I just wanted to put it here in case anyone had a similar issue.

### Severity

annoyance

1. basic project setup
2. beforeLoad using type beforeLoadEvent in +layout.ts in routes
3. generate types
4. `ERROR: typescript: 'beforeLoad' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.`

### Reproduction

_No response_

sjcobb2022:
Fixes #718

This should fix things properly now. Sorry for the inconvenience this was a stupid oversight on my part.

### DIFF
--- a/packages/houdini-svelte/src/plugin/codegen/routes/index.ts
+++ b/packages/houdini-svelte/src/plugin/codegen/routes/index.ts
@@ -137,8 +137,8 @@ export default async function svelteKitGenerator(
 				const functionImports =
 					functionImportsToBring.length > 0
 						? `\nimport type { ${functionImportsToBring.join(
-								', '
-						  )} } from '${houdiniRelative}/plugins/houdini-svelte/runtime/types';`
+							', '
+						)} } from '${houdiniRelative}/plugins/houdini-svelte/runtime/types';`
 						: ''
 
 				// mutate typeImports with our functionImports, layout/pageStores, $result, $input,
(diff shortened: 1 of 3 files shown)


packages/houdini-svelte/src/plugin/codegen/routes/index.ts
```diff
@@ -324,7 +324,7 @@ function internal_append_afterLoad(queries: OperationDefinitionNode[]) {
 function append_beforeLoad(beforeLoad: boolean, type: 'Layout' | 'Page') {
 	return beforeLoad
 		? `
-export type BeforeLoadEvent = PageLoadEvent;
+export type BeforeLoadEvent = ${type}LoadEvent;
```
AlecAivazis, line 327: Oh whoops! totally should have caught that myself. Anyway, thanks for fixing it 👍

AlecAivazis (approved):

PR #723, opened 2022-11-21
docs: add theme switcher

### LINKED ISSUE #634: Add light mode to docs
### Describe the feature

Hi, 

Dark theme burns my eyes so I can't read the documentation (unless reader mode which isn't really practical in this case - smaller code blocks, no syntax coloration...), it's a small detail but it makes it _really_ hard to learn houdini (which look awesome by the way). 

Usually projects docs are either light or light+dark but almost never just dark (I suppose because of this). 

Thank you!  

### Criticality

_No response_

gterras:
Fixes #634

Here is a basic implementation of a CSS theme switcher, a few notes and problems below, if you have any advice please don't hesitate!  

### ~~Preference lost on navigation~~

~~Apparently due to the specific way layouts are handled the binded theme variable `<ThemeSwitcher bind:ui_theme />` can't be persisted on (client) navigation.  The variable is set to `light` so it's easier to review but it's meant to be switched to `dark` when it's resolved (note that the CSS variables insert in `<svelte-head>` defaults to dark).~~  

### ~~No Persistence~~  

~~For the reason above and also because I wanted your input before going further, persistence is not implemented. It should be straightforward to set a cookie and get it in in a `load.server` function but maybe you will think of another way. On the other hand if `OS Default` choice is implemented (see below) it can be almost forgivable not to bother with persistence at all.~~  

### ~~Aria markup~~

~~The module markup is taken from MDN so it should be all right. I'm not fully aware of how aria works so there could be overlooks.~~ 

### Mechanics 

~~An icon button that open a panel containing two buttons. It could be interesting to implement an `OS default` third choice (and maybe default) once basics are reviewed.~~ 

The technique used is CSS variables injection in `<svelte-head>` from `_page.svelte` layout. Doing so isn't really academic but it gives the advantage of being able to use the variables in static stylesheets (like it was needed for scrollbar styling  and a very slight code styling).  

### Style 

So a bit of refactoring was needed on colors before going started. There are a few micro-differences on the dark theme because of this (mostly on "deep dive" blocks but I can point them precisely). 

For the light theme I've tried to stay as simple as possible. Maybe we can go further afterwards notably for the code blocks.

The homepage is untouched and has no theme switcher. All the other pages should be covered, as well as search modal and mobile mode. 

**Couldn't find test for site.**
**Couldn't find `integration` folder** 


site/src/components/ThemeSwitcher.svelte
```diff
@@ -0,0 +1,127 @@
...
+	const toggleMenu = () => (menuOpen = !menuOpen)
+
+	const setTheme = (themeName) => {
+		ui_theme = themeName
+		menuOpen = false
+	}
+</script>
+
+<div class="theme-switcher">
+	<button
```
AlecAivazis, line 32: How would you feel if we just had the button toggle the state instead of opening up a dropdown?

AlecAivazis (commented):
I really appreciate all the work @gterras! A few of the colors look a little off to me but I'll wait until the PR comes out of draft to give it a more thorough review

gterras:
> I really appreciate all the work @gterras! but I'll wait until the PR comes out of draft to give it a more thorough review

That's the least I can do! 

>  A few of the colors look a little off to me

Colors aren't totally polished yet but I was thinking you have the final word on it anyway. I've tried to declare the least amount of CSS variables possible so that it's easy to enhance. 

> until the PR comes out of draft to give it a more thorough review

I would have loved your feedback on `Preference lost on navigation` and `No Persistence` to go further.

site/src/components/ThemeSwitcher.svelte
gterras: Indeed as it is right now a toggle would be more fit, I let it this way for now so it's easier to integrate the potential third choice `OS Default` (see `No Persistence` and  `Mechanics`). If you are up for a third choice the dropdown would be the natural fit (and if you're not, reverting to a toggle would allow a bit of cleanup in the `<ThemeSwitcher>` data structure. (getting rid of the `Object.keys` and such).

AlecAivazis:
> I would have loved your feedback on Preference lost on navigation and No Persistence to go further.

Yea, sorry I didn't have any super concrete - things have been very busy at work / home. I think using a cookie is probably the best approach. Do you know of a way to get the OS default from the browser? Since the default value isn't available on SSR, it'd be nice if there was a way to access it...

site/src/components/ThemeSwitcher.svelte
gterras: I think  3-states toggle is quite a challenge to properly setup and ultimately offers a confusing UX. As a user I will choose with no hesitation a dropdown over a "click and see what happen, then maybe click again" experience. 
Since OS default seems to be tricky to implement (UI, SSR, icon) maybe we should just stick to dark by default and have a simple toggle to switch to light, what do you think?

gterras: @AlecAivazis Do you have any feedback on the above and the other comment? Thanks!

AlecAivazis: Sorry I went quiet, I've had a lot on my plate over the holidays. I think what you suggested is a reasonable compromise. Btw, if I don't respond in a few days on Github, feel free to ping me on discord. I'm more likely to see that message than a notification here

524c:
Hey @gterras and @AlecAivazis, 

Maybe I can help with that. What do you think of this site as a reference https://flowbite-svelte.com?

AlecAivazis:
Oh that seems to work really well! I like having the toggle as a click 👍

gterras:
Hi @AlecAivazis a working implementation, notable points: 

- toggle instead of menu 
- persistence through cookie (added a `site/src/routes/+layout.server.js` for this purpose)
- moved `<svelte:head>` to `ThemeSwitcher` component for a truly isolated component 
- switched to boolean for theme value (felt cleaner but revertable if you prefer string) 
- data structure cleanup in `ThemeSwitcher` (note that `themes` is feels reversed it's because you actually need to display non-current theme icon and label)

Ready for review !

AlecAivazis:
Just wanna thank you again for putting so much energy into this. I finally found some time to pull it down and ran into a small problem. My system is configured to use dark mode and then I click on the toggle to show the light mode UI, navigating using the sidebar takes me to a dark mode ui.

I do have a few small design tweaks but I figured I could just push them up to your branch if that's okay with you.

PR #727, opened 2022-11-24
limits - lib reporter

jycouet:
_Internal_
Just to make sure we control the bundle size.

Will break the build if we reach limits.
(It's ok to increase limits on adding features & debuging, we just want have everything under control)


jycouet:
By doing
```
cd e2e/sveltekit
pnpm build
pnpm libReporter
```

You will see this graph 👇

image

AlecAivazis (approved):

PR #732, opened 2022-11-30
fix: 🐛 do not delay app start by 2 schema pulls

janvotava:
Because of the sync nature of Houdini Watch Schema plugin start it was doing pull, afterwards one sync [sleep](https://github.com/HoudiniGraphql/houdini/blob/main/packages/houdini/src/vite/schema.ts#L71), then another pull and [sync sleep](https://github.com/HoudiniGraphql/houdini/blob/main/packages/houdini/src/vite/schema.ts#L54) and a then finally it switched to being [async](https://github.com/HoudiniGraphql/houdini/blob/main/packages/houdini/src/vite/schema.ts#L57) and application could start. In case that `schemaPullInterval` was configured for example to 60 seconds it would take =~ 120 seconds for an application to start.


janvotava:
Actually this won't wait for a first pull properly 🤔. Will fix that first.

PR #738, opened 2022-12-02
Feat - default client

### LINKED ISSUE #734: Use env in houdini.config.js
### Describe the bug

I'm trying to import my GraphQL endpoint in the houdini.config.js file via the .env file, but I unable to get it working. 
I have tried the following methods:

```
import { $env } from '$env/dynamic/public';
apiUrl: $env.PUBLIC_GRAPHQL_ENDPOINT
```

```
import { PUBLIC_GRAPHQL_ENDPOINT } from '$env/static/public';
apiUrl: PUBLIC_GRAPHQL_ENDPOINT
```

`apiUrl: process.env.PUBLIC_GRAPHQL_ENDPOINT`

### Severity

Add variable to .env and try to import it houdini.config.js

### Reproduction

_No response_

### LINKED ISSUE #735: client config default value
### Describe the feature

https://github.com/HoudiniGraphql/houdini/discussions/459#discussioncomment-3442491

### Criticality

nice to have

jycouet:
Fixes #734, fixes #735


JacobArrow:
Looks promising. Great job!

packages/houdini-svelte/src/plugin/index.ts
```diff
@@ -122,12 +122,15 @@ export const error = svelteKitError
 			})
 		}
 
-		if (!cfg.configFile.plugins?.['houdini-svelte'].client) {
+		const cfgPlugin = plugin_config(cfg)
```
AlecAivazis, line 125: I'm pretty sure this is a breaking change. The old config used to work if the user provided an extension or not. Can we support both situations here? I guess it's really 4 different cases

AlecAivazis (changes requested):

packages/houdini-svelte/src/plugin/index.ts
AlecAivazis: Also, while you're here could you use the async versions of the functions?

AlecAivazis (approved):

PR #740, opened 2022-12-03
upgrade node to 18

524c:
Fixes #TICKET

Node.js 18 has been LTS now

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,7 +33,7 @@ jobs:
             - name: Install Node.js
               uses: actions/setup-node@v3
               with:
-                node-version: 16.17.0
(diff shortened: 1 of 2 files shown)


AlecAivazis:
Thanks for submitting this @524c. Is there a problem with using the older version? I was doing it somewhat on purpose in the tests to make sure that we supported 16.x

524c:
> Thanks for submitting this @524c. Is there a problem with using the older version? I was doing it somewhat on purpose in the tests to make sure that we supported 16.x

Ah ok, I understand.  No problem at all, it was just proactive 😁

AlecAivazis:
No problem at all! Appreciate the effort 👍

PR #747, opened 2022-12-07
Improve support for unions and interfaces

### LINKED ISSUE #651: Arrays of union types with nested fields with same name but different type causes "could not merge" error
### Describe the bug

With a schema that has an array with a union type, if there is a field with the same name with a different type in the union types, it causes a "could not merge" warning error during codegen and does not fill the data properly in the query results.

The same query works when run directly against the GraphQL server.

The "could not merge" error seems to be from a check when building the query selection code that makes sure that each type is the same.
https://github.com/HoudiniGraphql/houdini/blob/main/packages/houdini/src/codegen/generators/artifacts/utils.ts#L46-L56

I believe the deepMerge call that fails comes from here https://github.com/HoudiniGraphql/houdini/blob/main/packages/houdini/src/codegen/generators/artifacts/selection.ts#L49

The workaround is to alias the field names in the query fragments so no fragment in the union has the same field name with different types. But then you have to update all of your components to use the new field name and generally remember that you have aliased that field because it doesn't match the schema.

I'm interested to attempt to help create a fix, but it seems like a tricky problem.

### Severity

Given a schema like this:

```graphql
interface Content {
  name: String!
}

type Page implements Content {
  id: ID!
  images: PageImages!
  name: String!
}

type PageImages {
  footer: String
  header: String
}

type Article implements Content {
  id: ID!
  images: ArticleImages!
  name: String!
}

type ArticleImages {
  author: String
  header: String
}

type Query {
  allContent: [Content!]!
}
```

And then this query:

```graphql
query AllContent {
    allContent {
        name
        ... on Page {
            images {
                header
                footer
            }
        }
        ... on Article {
            images {
                header
                author
            }
        }
    }
}
```

### Reproduction

https://github.com/rmarscher/houdinigraphql-fragment-merge-repro

### LINKED ISSUE #731: subscribe caching collide variables/queries from different operations
### Describe the bug

Two different operations interfere with each other and doing a store reset after a comparison of changes in variables, which did not occur.

### Severity

I'm working on a mini project and I'll update this report soon. For now I made two videos to demonstrate what happens.
After an initial investigation it seems to me that the problem is in the logic of the cache.

The videos below are using the current version 0.17.13 and a simple Svelte project page.

Each button does a different query. The expected behavior is that the result would be persisted in the store, but one operation interferes with the other.
Fisrt video without any changes in Houdini code:

https://user-images.githubusercontent.com/527641/204096718-beffa950-008d-42d0-bd72-a1ba878d1eec.mp4

This second video is after commenting the refreshSubscription function. That way I managed to keep the data of both operations.

Note: in this quick test I made direct changes to the volatile runtime files.
$houdini/plugins/houdini-svelte/runtime/stores/query.js
```
  refreshSubscription(newVariables) {
    /*
    const cache = getCache();
    if (this.subscriptionSpec) {
      cache.unsubscribe(this.subscriptionSpec, this.lastVariables || {});
    }
    this.subscriptionSpec = {
      rootType: this.artifact.rootType,
      selection: this.artifact.selection,
      variables: () => newVariables,
      set: (newValue) => this.store.update((s) => ({ ...s, data: newValue }))
    };
    cache.subscribe(this.subscriptionSpec, newVariables);
    this.lastVariables = newVariables;*/
  }
```
Result after this "code change":

https://user-images.githubusercontent.com/527641/204097169-08e1d188-492d-4abc-b68d-c7fa8b47c285.mp4

+page.svelte
```
<script lang="ts">
  import AllUsers from '$lib/components/utils/AllUsers.svelte';
  import UserPk from '$lib/components/utils/UserPk.svelte';

  export let data: any;

  $: ({ userPk } = data);
</script>

<br />

<AllUsers />
<br />

<UserPk pk={1} />
<br />
```

Component UserPk.svelte
```
<script lang="ts">
  import { G_userPk } from '$houdini';
  import { Button } from 'flowbite-svelte';

  export let pk = 0;
  const startColor = 'mt-2 flex flex-wrap text-red-500';
  const normalColor = 'mt-2 flex flex-wrap text-blue-500';
  $: color = startColor;

  export function userPkVariables({ props }: { props: { pk?: number } }) {
    return {
      pk: props.pk
    };
  }

  async function userPk() {
    color = startColor;
    await G_userPk.fetch({ variables: { pk } });
    setTimeout(() => {
      color = normalColor;
    }, 1000);
  }
</script>

<Button on:click={userPk}>userPk</Button>
<br />

{#if $G_userPk?.data}
  <pre class={color}>G_userPk: {JSON.stringify($G_userPk?.data, null, 2)}</pre>
  <br />
{/if}
```

Component AllUsers.svelte
```
<script lang="ts">
  import { G_allUsers } from '$houdini';
  import { Button } from 'flowbite-svelte';
  const startColor = 'mt-2 flex flex-wrap text-red-500';
  const normalColor = 'mt-2 flex flex-wrap text-blue-500';

  $: color = startColor;

  async function allUsers() {
    color = startColor;
    await G_allUsers.fetch();
    setTimeout(() => {
      color = normalColor;
    }, 1000);
  }
</script>

<Button on:click={allUsers}>allUsers</Button>
<br />

{#if $G_allUsers?.data}
  <pre class={color}>G_allUsers: {JSON.stringify($G_allUsers?.data, null, 2)}</pre>
  <br />
{/if}
```

### Reproduction

_No response_

AlecAivazis:
Fixes #731, fixes #651

This PR fixes an issue in the generated artifacts that prevented different selections from being retrieved depending on the concrete type of the record


AlecAivazis:
I'm going to merge this and give people a few more days to confirm if it works for them before publishing

PR #752, opened 2022-12-09
prefix function

### LINKED ISSUE #597: Variable function type collides with exported function name
### Describe the feature

At the moment, it's a bit awkward to use the variable function type definitions since they match the same name as the exported function. This forces the user to do something like:

```typescript
import { FooVariables as Variables } from '$houdini'

export const FooVariables: Variables = () => { ...}
```

One option is to change the name of the generated type to avoid the collision but maybe there are other options?

### LINKED ISSUE #751: Invalid export 'houdini_load'
### Describe the bug

I tried to play around with the example and i think one of the last SvelteKit changes broke houdini. 
Kit now disallows any non kit exports from +page.js

``` bash
Invalid export 'houdini_load' in src/routes/[filter]/+page.ts (valid exports are load, prerender, csr, ssr, trailingSlash, or anything with a '_' prefix)
```

### Severity

blocking all usage of Houdini

1. Clone the reproduction
2. Run `pnpm dev`
3. View the error in the console

### Reproduction

https://github.com/pixelmund/houdini-invalid-export-repro

jycouet:
Fixes #751, closes #597 

Since SvelteKit `1.0.0-next.573`, we are not allowed to export non-Kit functions from `+page.js`, or it has to start with "`_`".


.changeset/swift-bags-dress.md
```diff
@@ -0,0 +1,5 @@
+---
+'houdini-svelte': patch
```
AlecAivazis, line 2: This needs to be a minor

packages/houdini-svelte/src/plugin/codegen/routes/kit.test.ts
```diff
@@ -780,14 +764,12 @@ test('generates types for page beforeLoad', async function () {
 		export type PageData = Expand<Expand<Omit<PageParentData, keyof PageParentData & EnsureDefined<PageServerData>> & OptionalUnion<EnsureDefined<PageParentData & EnsureDefined<PageServerData>>>> & {
 		    MyPageLoad1Query: MyPageLoad1QueryStore
 		    MyPageLoad2Query: MyPageLoad2QueryStore
-		} & BeforeLoadReturn>;
```
AlecAivazis, line 783: This seems strange - don't we need to include the beforeLoad return values?

packages/houdini-svelte/src/plugin/codegen/routes/kit.test.ts
```diff
@@ -969,61 +938,48 @@ test('generates types for page afterLoad', async function () {
...
-type AfterLoadData = {
-    MyPageLoad1Query: MyPageLoad1Query$result
-    MyPageLoad2Query: MyPageLoad2Query$result
-};
+		export type PageData = Expand<Expand<Omit<PageParentData, keyof PageParentData & EnsureDefined<PageServerData>> & OptionalUnion<EnsureDefined<PageParentData & EnsureDefined<PageServerData>>>> & {
+		    MyPageLoad1Query: MyPageLoad1QueryStore
+		    MyPageLoad2Query: MyPageLoad2QueryStore
+		}>;
 
-export type AfterLoadEvent = {
```
AlecAivazis, line 1021: Where did this definition go?

AlecAivazis (changes requested):

packages/houdini-svelte/src/plugin/codegen/routes/kit.test.ts
jycouet: Test was outdated. I think it's better now.

packages/houdini-svelte/src/plugin/codegen/routes/kit.test.ts
jycouet: Test was outdated. I think it's better now.

sjcobb2022:
Just a quick comment. Wouldn't this effect #597?

Having the type without the _ prefix prevents the collision.

jycouet:
> Just a quick comment. Wouldn't this effect/fix #597?
> 
> Having the type without the _ prefix prevents the collision.

Yes, you are right. We will have no more collisions with this fix.

You will be able to do:
```ts
import { FooVariables } from '$houdini'

export const _FooVariables: FooVariables = () => { ...}
```

---

_Just FYI, I will probably continue to use:_
```ts
import { FooVariables as V } from '$houdini'

export const _FooVariables: V = () => { ...}
```
_To have as less "FooVariables" as possible, but I guess that it's only a preference_

AlecAivazis (approved):

PR #760, opened 2022-12-13
new: plugin global store

jycouet:
Move global stores to a separate package `houdini-plugin-svelte-global-stores`


packages/houdini-svelte/src/plugin/codegen/stores/query.ts
```diff
@@ -62,9 +61,7 @@ export async function load_${artifactName}(params) {
 	}
 }
 
-export const ${globalStoreName} = new ${storeName}()
-
-export default ${globalStoreName}
+export default new ${storeName}()
```
AlecAivazis, line 64: Let's remove this too while we're at it

AlecAivazis (changes requested):

site/src/routes/api/graphql/+page.svx
```diff
@@ -129,9 +129,11 @@ Houdini generates runtime representations of every enum in your schema. They can
...
 
 ```javascript
-import { MyEnum, GQL_MyMutation } from '$houdini'
+import { MyEnum, MyMutationStore } from '$houdini'
+
+const update = new MyMutationStore()
 
 function onClick() {
-	GQL_MyMutation.mutate({ variables: { input: MyEnum.Value1 } })
+	update.mutate({ variables: { input: MyEnum.Value1 } })
```
AlecAivazis, line 137: These examples make me think we should make `mutate` a static method? it will still return the value if you want to use it.... what do you think?

jycouet: But we would still need `$update.isFetching`
Maybe we can add it in the example to show the purpose of the two steps?

packages/houdini-svelte/src/plugin/codegen/stores/query.ts
jycouet: 👍 Done

jycouet: Actually not done yet as in transforms we used `default` export.
I'll have a look

jycouet: `import GQL_TestQuery from "$houdini/plugins/houdini-svelte/stores/TestQuery";`
is used in 2 places (load): 
- `"artifact": GQL_MyQuery2.artifact`, so I switched to use: `"artifact": _MyQuery2Artifact`
- `const store1 = GQL_MyQuery1;`, so we should have `new MyQuery1()`

In the end `export function store_import` function should not exist as it's using the default!

Maybe we can have a look together later.

site/src/routes/api/graphql/+page.svx
jycouet: No new MyMutationStore() but 

```
const foo = graphql`

`

foo.mutate()
```

packages/houdini-svelte/src/plugin/codegen/stores/query.ts
jycouet: @AlecAivazis check this plz 🚀

packages/houdini-svelte/src/plugin/codegen/stores/fragment.ts
```diff
@@ -68,8 +63,6 @@ export declare class ${storeName} extends ${store_class}<${_data}, {}> {
 	}
 }
 
-export const ${globalStoreName}: ${storeName}
-
 export declare const load_${artifactName}: (params: QueryStoreFetchParams<${_data}, {}>) => Promise<${storeName}>
 
 export default ${storeName}
```
AlecAivazis, line 68: Reminder to remove this and the other default exports in the typedefs 💥

AlecAivazis (changes requested):

packages/houdini-svelte/src/plugin/codegen/stores/fragment.ts
jycouet: Yep, good catch ✅

AlecAivazis (approved):

PR #767, opened 2022-12-15
Add support for expr satisfies Type expressions

### LINKED ISSUE #765: Houdini doesn't work with new TS v4.9 satisfies operator
### Describe the bug

The Sveltekit generate TS with new `satisfies operator` since this [PR](https://github.com/sveltejs/kit/pull/8001).

I updated my code `hooks.server.ts` to use this, [like this example](https://kit.svelte.dev/docs/hooks#server-hooks), and I get an error.

The output after running the `houdini generate -v` command and updating my `hooks.server.ts` to use this new operator:
```
❌ Encountered error in src/hooks.server.ts
    at file:///devel/.local/share/pnpm/global/5/.pnpm/houdini@0.18.1/node_modules/houdini/build/cmd-esm/index.js:76213:15
    at async Promise.all (index 2)
    at async collectDocuments (file:///devel/.local/share/pnpm/global/5/.pnpm/houdini@0.18.1/node_modules/houdini/build/cmd-esm/index.js:76188:3)
    at async compile (file:///devel/.local/share/pnpm/global/5/.pnpm/houdini@0.18.1/node_modules/houdini/build/cmd-esm/index.js:76080:21)
    at async Command2.generate 
(file:///devel/.local/share/pnpm/global/5/.pnpm/houdini@0.18.1/node_modules/houdini/build/cmd-esm/index.js:76375:5)
```

### Severity

It is possible to use `houdini/e2e/sveltekit` to reproduce the error, and then:

update the src/hooks.server.ts with with this code:
```typescript
import { setSession } from '$houdini';
import type { Handle } from '@sveltejs/kit';

export const handle = (async ({ event, resolve }) => {
  // set the session information for this event
  setSession(event, { user: { token: '1234-Houdini-Token-5678' } });

  // pass the event onto the default handle
  return await resolve(event);
}) satisfies Handle;
```

then run `houdini generate -v`

### Reproduction

_No response_

524c:
Fixes #765

### DIFF
--- a/.changeset/angry-teachers-design.md
+++ b/.changeset/angry-teachers-design.md
@@ -1,10 +1,10 @@
 ---
-'e2e-next': minor
-'sveltekit': minor
-'scripts': minor
-'houdini': minor
-'houdini-react': minor
-'houdini-svelte': minor
+'e2e-next': patch
+'sveltekit': patch
+'scripts': patch
+'houdini': patch
+'houdini-react': patch
+'houdini-svelte': patch
 ---
 
 Add support for expr satisfies TS expressions
--- a/.changeset/angry-teachers-design.md
+++ b/.changeset/angry-teachers-design.md
@@ -1,0 +1,10 @@
+---
+'e2e-next': minor
+'sveltekit': minor
+'scripts': minor
+'houdini': minor
+'houdini-react': minor
+'houdini-svelte': minor
+---
+
+Add support for expr satisfies TS expressions
--- a/e2e/next/package.json
+++ b/e2e/next/package.json
@@ -25,6 +25,6 @@
 		"eslint-config-next": "13.0.0",
 		"houdini": "workspace:^",
 		"houdini-react": "workspace:^",
-		"typescript": "4.8.4"
+		"typescript": "^4.9"
 	}
 }
--- a/e2e/sveltekit/package.json
+++ b/e2e/sveltekit/package.json
@@ -39,7 +39,7 @@
     "svelte-check": "^2.2.6",
     "svelte-preprocess": "^4.10.1",
     "tslib": "^2.3.1",
-    "typescript": "~4.6.2",
+    "typescript": "^4.9",
     "vite": "^4.0.1",
     "vite-plugin-lib-reporter": "^0.0.6"
   }
--- a/e2e/sveltekit/src/hooks.server.ts
+++ b/e2e/sveltekit/src/hooks.server.ts
@@ -1,10 +1,10 @@
 import { setSession } from '$houdini';
 import type { Handle } from '@sveltejs/kit';
 
-export const handle: Handle = async ({ event, resolve }) => {
+export const handle = (async ({ event, resolve }) => {
   // set the session information for this event
   setSession(event, { user: { token: '1234-Houdini-Token-5678' } });
 
   // pass the event onto the default handle
   return await resolve(event);
-};
+}) satisfies Handle;
--- a/example/package.json
+++ b/example/package.json
@@ -21,7 +21,7 @@
 		"svelte": "^3.55.0",
 		"svelte-preprocess": "^4.10.1",
 		"tslib": "^2.3.1",
-		"typescript": "^4.8.4",
+		"typescript": "^4.9",
 		"vite": "^4.0.1",
 		"ws": "^8.8.1"
 	},
--- a/package.json
+++ b/package.json
@@ -43,6 +43,6 @@
 		"fs-extra": "^10.1.0",
 		"jest-snapshot": "^29.1.2",
 		"memfs": "^3.4.7",
-		"recast": "^0.21.5"
+		"recast": "^0.23.1"
 	}
 }
--- a/packages/_scripts/package.json
+++ b/packages/_scripts/package.json
@@ -16,6 +16,6 @@
 		"glob": "^8.0.3",
 		"rollup": "^3.7.4",
 		"rollup-plugin-typescript2": "^0.34.0",
-		"typescript": "^4.8.4"
+		"typescript": "^4.9"
 	}
 }
--- a/packages/houdini-react/package.json
+++ b/packages/houdini-react/package.json
@@ -32,7 +32,7 @@
         "estree-walker": "^3.0.1",
         "graphql": "^15.8.0",
         "houdini": "workspace:^",
-        "recast": "^0.21.5"
+        "recast": "^0.23.1"
     },
     "files": [
         "build"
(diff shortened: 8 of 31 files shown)


524c:
I'm investigating the error.

AlecAivazis:
Awesome - thanks for pushing this through! Mind adding a changeset?

524c:
I need to do an update vitest snapshot because for some reason the project followed my local vscode setup (indentation spacing, semicolons at the end of each line of code, etc...).
@AlecAivazis  or @jycouet, how do you think we should handle this?

I thought that my vscode would follow the project's setup and not something "global" in my environment. Other than that everything looks normal.

houdini-vitest-snapshot-2
houdini-vitest-snapshot-1

524c:
> Awesome - thanks for pushing this through! Mind adding a changeset?

Hey @AlecAivazis
I will create. First I would like help, please, with this question I commented on vitest.

AlecAivazis:
Huh that's strange. We should definitely try to figure out why the snapshot isn't respecting the project setup. Do you have a `.vscode` directory locally?

524c:
> h that's strange. We should definitely try to figure out why the snapshot isn't respecting the project setup. Do you have a `.vscode` directory locally?

Not in this fork of Houdini 🫤

524c:
I believe it could be the following:
Maybe in your project you have vscode user settings, which don't exist in the project's local setup. So in my environment prevailed the configurations that I also have related to the user. But this is just a suspicion.

524c:
When you have time, if you can, could you run the tests again and a new snapshot to see if it changes again?

524c:
> When you have time, if you can, could you run the tests again and a new snapshot to see if it changes again, using this fork?

jycouet:
FYI, I'm looking at this format thing right now @524c

jycouet:
I think that it's related to: https://github.com/benjamn/recast/pull/1247
_Quote of the important part 👇_
> However, many/most TypeScript programmers would probably expect to see semicolons here instead of commas/nothing. We could make this configurable, but I'm not sure it's even controversial, so I'll wait for someone to ask.

This is part of `"recast": "^0.23.1"` (The version you bumped to), it's a "next" version, not the official one.
In `"recast": "^0.22.0"` "latest", this is not there yet, so snapshots don't change.

Maybe you can downgrade to latest?
Changing the format of generated code might be a thing to do when the lib will be officially released with it?

Let me know what do you think

524c:
Perfect, thanks for finding this.
I will downgrade and test everything again.

524c:
I forget things very quickly,but I remembered that I tried to use "recast": "^0.22.0", but it turns out that the current 0.22.0 depends on "ast-types": "0.15.2", which in turn does not support the new typescript operator.

I believe it will be better to wait until they release the "recast": "^0.23.1" as final.

AlecAivazis:
Imo as long as we understand the change in format and that it's not because of some setup in your config, I think using the next version that supports it seems like a reasonable thing to do

524c:
FYI, @jycouet and @AlecAivazis,
With "recast": "^0.22.0" i get erros like this (because the "ast-types": "0.15.2"):
```
houdini:typedefs:   Type 'TaggedTemplateExpression' is not assignable to type 'ExpressionKind | DeclarationKind'.
houdini:typedefs:     Type 'import("/mnt/data/projects/houdini/node_modules/.pnpm/ast-types@0.16.1/node_modules/ast-types/lib/gen/namedTypes").namedTypes.TaggedTemplateExpression' is not assignable to type 'import("/mnt/data/projects/houdini/node_modules/.pnpm/ast-types@0.15.2/node_modules/ast-types/gen/namedTypes").namedTypes.TaggedTemplateExpression'.
houdini:typedefs: /mnt/data/projects/houdini/packages/houdini/src/codegen/utils/moduleExport.ts (24,50): Argument of type 'ExpressionKind' is not assignable to parameter of type 'ExpressionKind | null | undefined'.
```

524c:
> Imo as long as we understand the change in format and that it's not because of some setup in your config, I think using the next version that supports it seems like a reasonable thing to do

At least it's nothing with vscode 🙏

AlecAivazis:
exactly. and looking closer at the version, I dont think this is a pre-release version or something like that.  It seems like you just installed the latest version published on npm 👍

.changeset/angry-teachers-design.md
```diff
@@ -0,0 +1,10 @@
+---
+'e2e-next': minor
+'sveltekit': minor
+'scripts': minor
+'houdini': minor
+'houdini-react': minor
+'houdini-svelte': minor
```
AlecAivazis, line 7: By the way, these are supposed to be `patch`. `Minor` is reserved for changes that add new features (or break the API while we're in a 0.x version). I'm going to push up the tweak but just wanted to mention it for future PRs

AlecAivazis: Oh, damn. I can't push up the fix because these are on your `main` brach. Mind changing this to 

```
---
'houdini': patch
'houdini-svelte': patch
---

Add support for expr satisfies TS expressions
```

524c:
> exactly. and looking closer at the version, I dont think this is a pre-release version or something like that. It seems like you just installed the latest version published on npm 👍

Yes, I installed via npm, although it is not listed. But as I saw on the github of the project that there was a more recent one that deals with the problem (0.23.1), i decided to try.

.changeset/angry-teachers-design.md
524c: > By the way, these are supposed to be `patch`. `Minor` is reserved for changes that add new features (or break the API while we're in a 0.x version). I'm going to push up the tweak but just wanted to mention it for future PRs

you're right, i'll fix it.

AlecAivazis:
> Yes, I installed via npm, although it is not listed. But as I saw on the github of the project that there was a more recent one that deals with the problem (0.23.1), i decided to try.

Oh, it _is_ a next version. hmm okay. I'm still not totally against it but i want to investigate it a little more

.changeset/angry-teachers-design.md
524c: No problem, I already changed.

AlecAivazis (approved):
Okay, thinking about this some more I think it's probably the right move to merge this since people will be using `satisfies` a lot in 1.0. If we run into issues we can resolve it then

524c:
> Okay, thinking about this some more I think it's probably the right move to merge this since people will be using `satisfies` a lot in 1.0. If we run into issues we can resolve it then

Yes, with the release of `1.0` things got a little hectic and a lot of people updating their stuff.

PR #768, opened 2022-12-15
Fix generated types referencing onError, afterLoad and beforeLoad

thokra:
Fixes an issue with `AfterLoadReturn`, `BeforeLoadReturn` and `OnErrorReturn` types. 

### DIFF
--- a/packages/houdini-svelte/src/plugin/codegen/routes/index.ts
+++ b/packages/houdini-svelte/src/plugin/codegen/routes/index.ts
@@ -297,7 +297,7 @@ function append_afterLoad(
 ) {
 	return afterLoad
 		? `
-type AfterLoadReturn = Awaited<ReturnType<typeof import('./+${type.toLowerCase()}').afterLoad>>;
+type AfterLoadReturn = Awaited<ReturnType<typeof import('./+${type.toLowerCase()}')._houdini_afterLoad>>;
 type AfterLoadData = {
 	${internal_append_afterLoad(queries)}
 };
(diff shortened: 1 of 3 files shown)


AlecAivazis:
🤦 thanks for fixing this!

packages/houdini-svelte/src/plugin/codegen/routes/index.ts
```diff
@@ -297,7 +297,7 @@ function append_afterLoad(
 ) {
 	return afterLoad
 		? `
-type AfterLoadReturn = Awaited<ReturnType<typeof import('./+${type.toLowerCase()}').afterLoad>>;
+type AfterLoadReturn = Awaited<ReturnType<typeof import('./+${type.toLowerCase()}')._houdini_afterLoad>>;
```
AlecAivazis, line 300: Mind changing these to use the values exported from [here](https://github.com/HoudiniGraphql/houdini/blob/d1fcc479791c3477cee4b5fd006c44cd9aab60b9/packages/houdini-svelte/src/plugin/naming.ts)?

AlecAivazis (changes requested):

packages/houdini-svelte/src/plugin/codegen/routes/index.ts
524c: > Mind changing these to use the values exported from [here](https://github.com/HoudiniGraphql/houdini/blob/d1fcc479791c3477cee4b5fd006c44cd9aab60b9/packages/houdini-svelte/src/plugin/naming.ts)?

ok, I'm looking at this right now.

AlecAivazis: I think you got confused by a notification @524c 😅

524c: I just got my browser tabs mixed up, it has nothing to do with my PR, sorry 😆

AlecAivazis (approved):

PR #771, opened 2022-12-15
tweak doc styling

ronkeiser:
Fixes #TICKET


site/src/components/ThemeSwitcher.svelte
```diff
@@ -54,6 +55,7 @@
 				--graphql-explained: #8d005f;
 				--graphql-explained-background: var(--discreet);
 				--text_highlight: #a1c5f8;
+				--text_highlight-background:  red;
```
AlecAivazis, line 58: Is this intentional?

AlecAivazis (approved):
Thanks for this! 🙌

PR #776, opened 2022-12-17
Graphql as function

AlecAivazis:
This PR adds the ability to use `graphql` as a function (as well as a template tag). The svelte plugin uses a new hook to provide overloaded definitions of the tag to map query strings to the appropriate store value. This means that users who use the function no longer have to explicitly type their values.

- [ ] ~Update the first line to point to the ticket that this PR fixes~

### What's Left
- [x] confirm fragment is required to pass into fragment
- [x] fragment and paginated fragments need to pass the stores through


jycouet (changes requested):
Removing a lot of `: xxxStore` feels so good!
Great stuff 🥳

PR #789, opened 2022-12-24
Infer query inputs from route params

### LINKED ISSUE #372: Infer inline query variables from route parameters
If a route's path contains parameters that match a query's variables, the user shouldn't have to define a variable function if its just going to do `({params}) => params`.  

Questions:
- How do we handle massaging custom scalars?

### LINKED ISSUE #468: Allow users to toggle between typescript and jsdocs examples

AlecAivazis:
closes #372, closes #468

This PR updates Houdini's automatic load generation to use url parameters as query inputs when possible. For example, if a route is defined at `src/routes/users/[id]` and a query looks like this:

```graphql
query UserInfo($id: Int!) {
    user(id: $id) {
        ...
    }
}
```

Then the `Variable` function is not required.  That being said, users can still define a `Variable` function for custom logic if they want. 


.changeset/calm-brooms-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+'houdini-svelte': minor
+---
+
+Removed this.error and this.redirect from function variables
```
AlecAivazis, line 5: Still not really sure if this is worth doing or not. It simplifies the generated code significantly but does introduce yet _another_ breaking change so soon after `0.20`

packages/houdini-svelte/src/plugin/transforms/kit/load.ts
```diff
@@ -517,3 +492,150 @@ function unexported_data_error(filepath: string) {
 		description: `This is not allowed in a route since it would conflict with Houdini's generated code`,
 	}
 }
+
+function variable_function_for_query(
```
jycouet, line 496: Could we also check for `url.searchParams.get('xxx');`
It's quite useful when you deal with filters

site/src/components/Toolbar.svelte
```diff
@@ -0,0 +1,185 @@
+<script>
+	import { Icon } from '~/components'
+
+	export let ui_theme = 0
+	export let lang = 'js'
+	$: other_lang = lang === 'js' ? 'ts' : 'js'
+	const full_name = {
```
jycouet, line 7: The button has the same behavior as the theme, but I wonder if the opposite is not clearer.
By displaying "The option you are in".

You are in js, you display js.
You are in dark, you display dark.

On the day we have 3 options on one button it will be also handled.

.changeset/calm-brooms-sit.md
jycouet: I think that it's also closer to kit way of doing, I would go for it. 👍

jycouet (commented):
Already looks great with variables.
But with docs on top, it's amazing.

site/src/components/Toolbar.svelte
AlecAivazis: Yea I was on the fence about this. If you think its better the other way that works for me 👍

PR #808, opened 2023-01-04
fix on:click on doc

FlymeDllVa:
A couple of corrections to the documentation

### DIFF
--- a/site/src/routes/guides/pagination/+page.svx
+++ b/site/src/routes/guides/pagination/+page.svx
@@ -140,7 +140,7 @@ For information about how to customize this behavior, check out the [Caching Dat
 ## Mutation Operations
 
 A paginated field can be marked as a potential target for a mutation operation by passing
-a `name` argument to the `@paginate` directive. For more information, see [this document](/api/mutation/inline#lists).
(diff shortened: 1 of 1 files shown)


site/src/routes/guides/pagination/+page.svx
```diff
@@ -140,7 +140,7 @@ For information about how to customize this behavior, check out the [Caching Dat
 ## Mutation Operations
 
 A paginated field can be marked as a potential target for a mutation operation by passing
-a `name` argument to the `@paginate` directive. For more information, see [this document](/api/mutation/inline#lists).
+a `name` argument to the `@paginate` directive. For more information, see [this document](/api/mutation#lists).
```
FlymeDllVa, line 143: I think the address is outdated?

AlecAivazis: oh yea, good catch!

AlecAivazis:
Nice! Thanks for submitting this. It'll be a little bit longer before the changes actually take effect on the main page since we are working on the `1.0` release and it includes doc changes that don't work in `0.19.x`

PR #814, opened 2023-01-07
Add RequestEvent in .mutate() to docs

fnimick:
Adds documentation as proposed in #812

### DIFF
--- a/site/src/routes/api/mutation/+page.svx
+++ b/site/src/routes/api/mutation/+page.svx
@@ -74,10 +74,13 @@ export const actions: Actions = {
             }
         `)
 
-        return await actionMutation.mutate({ name })
+        return await actionMutation.mutate({ name }, event)
     }
 }
 ```
+
+Please note that passing the `RequestEvent` through to the `mutate` call is necessary in order for
+SvelteKit's `fetch` request handling to work as expected: header forwarding, `handleFetch` hook, etc.
 
 ## Updating fields
 


AlecAivazis (approved):
Nice! Thanks for submitting this. It'll be a little bit longer before the changes actually take effect on the main page since we've paused updates. we are working on the 1.0 release and it includes doc changes that don't work in 0.19.x

xmlking:
can I pass SvelteKit `fetch` like this ?

```ts
xport const actions = {
	delete: async ({ request, fetch }) => {
		const formData = await request.formData();
		const id = formData.get('id')?.toString();
		if (!id) return fail(400, { id: 'id is required' });
		const variables = { id };

		try {
			const data = await deletePolicyStore.mutate(variables, {
				metadata: { authToken: 'token from session' },
				fetch
				// optimisticResponse: {}
			});

			const actionResult = data.delete_tz_policies_by_pk;
			if (!actionResult) return fail(400, { actionErrors: [{ message: 'data null' }] });

			return {
				actionResult
			};
		} catch (err) {
			if (isHttpError(err)) {
				throw error(err.status, err.body);
			}
			if (isAppError(err)) {
				throw error(500, err);
			}
			throw error(500, getAppError(err));
		}
	}
} satisfies Actions;

```

fnimick:
I believe that works - I would pass the whole `event` in though.
i.e.
```delete: async (event) => {
   const { request, fetch } = event;
   .... .mutate(variables, {metadata, ...event})
```

PR #815, opened 2023-01-08
Added graphql-sse in Subscriptions docs

FlymeDllVa:
Added graphql-sse in Subscriptions docs


site/src/routes/api/subscription/+page.svx
```diff
@@ -101,8 +101,27 @@ export default new HoudiniClient(requestHandler, socketClient)
 
 If your server supports the latest Server-Side events, you can use them to power your subscriptions:
 
+### Using `graphql-sse`
```
AlecAivazis, line 104: Do you know if mdsvex supports `####`? If so, I think it's probably better to treat `graphql-see` and `Without using third-party libraries` as children of the same section instead of siblings

AlecAivazis:
Hey @FlymeDllVa - mind opening this against the `next` branch? the API for subscription support changes pretty drastically but you should be able to adapt what you have pretty easily. Lemme know if you need any guidance

PR #820, opened 2023-01-13
Docker route detection

### LINKED ISSUE #817: Bug: SvelteKit build in node docker image fails - Cannot find module './$houdini'
### Describe the bug

Hello 👋 

When trying to build our SvelteKit application with a Dockerfile for deployment on the server, we currently receive an error when running the step `svelte-check`: "Error: Cannot find module './$houdini' or its corresponding type declarations. (ts)". Removing this step from the Dockerfile leads to an successful build, but then the application breaks during runtime: "Undefined property pageInfo" (using a pagination setup here).

After quite a lot of debugging, I figured out that the `houdini generate` command does not write all of the necessary files. Within the generated `$houdini` folder, the path `types/src/routes` is empty, where it should actually create all of the `.d.ts` files describing the available data for each page. On the other hand, the `$houdini/artifacts` folder is filled correctly. Why those files have not been written remains unknown. The standard SvelteKit type files (`.svelte-kit/types/src/routes/...`) are generated properly.

This issue only occurs when building the application in the environment of the node image. On my local system (macOS) there is no such issue. We make use of groups in the folder-based router (e.g. `src/routes/(protected)/articles/+page.svelte` and thought this might be an issue, but a route such as `src/routes/api/...` is not processed as well.

I hope my explanations above clarify the issue properly. If there is any other information required for debugging and analysis, please just let me know!

Any help in fixing this issue is very much appreciated 🙏  Thanks!

### Severity

blocking all usage of Houdini

1. Setup houdini with SvelteKit as per the official documentation.
2. Create a Dockerfile to build the product using the official node docker image.
3. Run `svelte-check` as part of the process to detect this error early on (otherwise it would come up at runtime).
4. Tested with all recent versions of houdini and houdini-svelte (0.18.0 -> 1.0.0-next.1).

### Reproduction

_No response_

AlecAivazis:
Fixes #817 

This PR fixes some build issues when building in docker

- [ ] ~If applicable, add a test that would fail without this fix~

### DIFF
--- a/packages/houdini-svelte/src/plugin/codegen/routes/index.ts
+++ b/packages/houdini-svelte/src/plugin/codegen/routes/index.ts
@@ -21,8 +21,6 @@ export default async function svelteKitGenerator(
 		return
 	}
 
-	console.log('generating')
-
 	await walk_routes(config, framework, {
 		async route({
 			dirpath,
--- a/packages/houdini-svelte/src/plugin/codegen/routes/index.ts
+++ b/packages/houdini-svelte/src/plugin/codegen/routes/index.ts
@@ -21,6 +21,8 @@ export default async function svelteKitGenerator(
 		return
 	}
 
+	console.log('generating')
+
 	await walk_routes(config, framework, {
 		async route({
 			dirpath,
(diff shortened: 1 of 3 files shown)


packages/houdini-svelte/src/plugin/codegen/routes/index.ts
```diff
@@ -21,6 +21,8 @@ export default async function svelteKitGenerator(
 		return
 	}
 
+	console.log('generating')
```
jycouet: I guess that you don't want to keep this?

AlecAivazis: it's already gone hahah

PR #839, opened 2023-01-21
✨ stale

### LINKED ISSUE #287: Add way to mark data as stale/invalid
### Discussed in https://github.com/HoudiniGraphql/houdini/discussions/286

<sup>Originally posted by **frederikhors** April 23, 2022</sup>
Is there an invalidate() api like https://formidable.com/open-source/urql/docs/graphcache/cache-updates/#invalidating-entities?</div>

jycouet:
Fixes #287

Adding `stale` mechanism. with 


e2e/sveltekit/src/routes/stale/add/+page.svelte
```diff
@@ -0,0 +1,65 @@
...
+  };
+
+  const markStale_type = async () => {
+    await addUser.mutate(null);
+    cache.markStale('UserNodes');
+  };
+
+  const markStale_type_field = async () => {
+    await addUser.mutate(null);
+    cache.markStale('UserNodes', 'totalCount');
```
AlecAivazis, line 24: Just a reminder for when we talk later:

This will need an `args` option Also, If we pass an object in as the first argument, users dont have to remember the order (plus it'd be more consistent with the rest of the API)

PR #841, opened 2023-01-22
Restructure programatic cache api

AlecAivazis:
This PR replaces the `get` and `set` methods in the imperative API with `read` and `write` which both take fragments:

```typescript
const user = cache.get('User', { id: 2 })

const data = user.read({
    fragment: graphql(`
        fragment UserReadInfo on User { 
            firstName
        }
    `)
})

user.write({
    fragment: graphql(`
        fragment UserWriteInfo on User { 
            firstName
        }
    `),
    data: { 
        firstName: "Another Name"
    }
})
```

While this is significantly more verbose than the previous api, there is no more of that `cache.setFieldType` complexity. the fragment contains all of the information we need to perform the write in a way that supports code splitting.


packages/houdini/src/runtime/public/types.ts
```diff
@@ -80,3 +94,35 @@ export type ListType<Def extends CacheTypeDef, Name extends ValidLists<Def>> = P
 	Def,
 	Def['lists'][Name]['types']
 >
+
+type FragmentDefinition<List, _Target> = List extends [infer Head, ...infer Rest]
```
AlecAivazis, line 98: Reminder to clean this up

packages/houdini-svelte/src/plugin/transforms/componentQuery.ts
```diff
@@ -224,7 +224,7 @@ export async function find_inline_queries(
 
 	// look for inline queries
 	await find_graphql(page.config, parsed, {
-		where(tag) {
+		where(tag, { parent }) {
```
AlecAivazis, line 227: Unused variable?

PR #852, opened 2023-01-26
:sparkles: NEW: global stores - no query stores by default

jycouet:
New config: `storesToGenerate` in `houdini-plugin-svelte-global-stores` plugin.
By default, query stores will not be generated, you can opt-in by config


.changeset/poor-hats-notice.md
```diff
@@ -0,0 +1,5 @@
+---
+'houdini-plugin-svelte-global-stores': patch
```
AlecAivazis, line 2: This should be `major` since it's a breaking change

e2e/sveltekit/houdini.config.js
```diff
@@ -22,7 +22,9 @@ const config = {
     }
   },
   plugins: {
-    'houdini-plugin-svelte-global-stores': {},
+    'houdini-plugin-svelte-global-stores': {
+      storesToGenerate: ['Query', 'Mutation', 'Subscription', 'Fragment']
```
AlecAivazis, line 26: can we do lowercase nouns for this? also i'd like to have an `all` for someone who doesn't want to be specific

AlecAivazis (changes requested):

PR #871, opened 2023-02-01
Merge 1.0 branch

### LINKED ISSUE #287: Add way to mark data as stale/invalid

### LINKED ISSUE #365: Network Layer
Issues like #285 and #290 make it pretty clear that Houdini needs some kind of network layer abstraction that simplifies the maintenance of a Client.

Whatever middleware model we choose, it needs to be able to support these situations :
- Query batching
- Global abort controller
- Automatic retry logic 
- Broadcast channels

Some examples of similar things in other libraries:
- [react-relay-network-layer](https://github.com/relay-tools/react-relay-network-modern)
- [urql exchange](https://formidable.com/open-source/urql/docs/advanced/authoring-exchanges/)
- [apollo links](https://www.apollographql.com/docs/react/api/link/introduction/)

### LINKED ISSUE #383: Optimistic Fetches
The computation for partial data relies on running into a field we've already seen before. We should be able to provide a hint to the cache when looking up the data so it can infer the result and potentially grab a larger partial result

a possible API:

```javascript
store.fetch({ variables: {}, optimistic: { ... } }) 
```

### LINKED ISSUE #466: Support for multiple endpoints
### Describe the feature

Due to reasons I need a custom `build.js` script that calls vite's `build()`, patches some files, moves the output to its target destination and creates a custom entry for the `handle` middleware.

It would be nice if I could integrate calling `houdini generate` into this file.

And in addition I'd love to have a script that pulls the schemas down (yes, there are mutliple...) with headers and all stuff. Basically I want to `import { writeSchema } from 'houdini/build/cmd/utils/writeSchema';`.

Currently all of this fails because of the `exports` field in the `package.json`.

### Criticality

nice to have

### LINKED ISSUE #670: Bidirectional cursor-based pagination
### Describe the feature

As discussed in #666, we should expand the current implementation of `@paginate` to support bi-directional pagination. 

I suspect that we should probably loosen the constraints on the current pagination behavior to not think of each direction as a distinct feature. The generated stores already have `loadPrevious` and `loadNext` so i think we should be able to just have both always working. The original constraint was in place because I was being a bit lazy about computing the initial page size but as the feature developed it ended up not being so necessary. It does help us a bit with the type information but again, if the store can just always do both then thats okay. 

One small hurdle is to figure out if we want to force users to provide page sizes for both directions. I think the best DX would be to give the user the freedom but not force them. That is to say, if they specify `first` and no `last` then we can just use the `first` value as the page size for reverse pagination (similarly for `last`). If they provide both, we should use the appropriate one for the appropriate direction. This will mean adding some bits to the generate artifact so we can persist the two values but that's relatively harmless.

### LINKED ISSUE #821: mutate should also return error similar to fetch
### Describe the feature

When configured with `quietQueryErrors: true`, **mutate** should also return error similar to **fetch**
if supported, I can do clean code to wrapping `GraphQLError` into business **error** that provide context to support team. 

```ts
import { env as dynPubEnv } from '$env/dynamic/public';
import { CachePolicy, DeletePolicyStore, order_by, SearchPoliciesStore } from '$houdini';
import { getAppError, isAppError, isHttpError, NotFoundError, PoliciyError } from '$lib/errors';
import { Logger } from '$lib/utils';
import { ZodError } from 'zod';
// import { getToken } from '@auth/core/jwt';
import { policiyDeleteSchema, policiySearchSchema } from '$lib/models/schema';
import { zx } from '$lib/zodsk';
import * as Sentry from '@sentry/svelte';
import { error, fail } from '@sveltejs/kit';
import type { GraphQLError } from 'graphql';
import assert from 'node:assert';
import type { Actions, PageServerLoad } from './$types';

assert.ok(dynPubEnv.PUBLIC_GRAPHQL_ENDPOINT, 'PUBLIC_GRAPHQL_ENDPOINT not configered');
assert.ok(dynPubEnv.PUBLIC_GRAPHQL_TOKEN, 'PUBLIC_GRAPHQL_TOKEN not configered');

const log = new Logger('policies.server');

const searchPoliciesStore = new SearchPoliciesStore();
const deletePolicyStore = new DeletePolicyStore();

export const load = (async (event) => {
	const { url, parent } = event;
	try {
		await parent(); // HINT: to make sure use session is valid

		const { limit, offset, subject_type, display_name } = zx.parseQuery(url.searchParams, policiySearchSchema);
		console.log(limit, offset, subject_type, display_name);
		const orderBy = [{ updated_at: order_by.desc_nulls_first }];
		const where = {
			deleted_at: { _is_null: true },
			subject_type: { _eq: subject_type },
			...(subject_type ? { subject_type: { _eq: subject_type } } : {}),
			...(display_name ? { display_name: { _like: `%${display_name}%` } } : {})
		};
		const variables = { where, limit, offset, orderBy };

		const { errors, data } = await searchPoliciesStore.fetch({
			event,
			blocking: true,
			policy: CachePolicy.CacheAndNetwork,
			metadata: { backendToken: 'token from TokenVault' },
			variables
		});

		if (errors) throw new PoliciyError({ name: 'SEARCH_POLICIY_ERROR', message: 'list policies api error', cause: errors[0] });
		if (!data) throw new NotFoundError('policies data is null');

		const {
			counts: { aggregate: count },
			tz_policies: policies
		} = data;

		// TODO: tune: This page will have cache for 5min
		// setHeaders({
		// 	'cache-control': 'public, max-age=300'
		// });

		return { count, policies };
	} catch (err) {
		log.debug('policies:actions:load:error:', err);

		if (err instanceof NotFoundError && err.name === 'NOT_FOUND_ERROR') {
			throw error(404, { message: err.message });
		} else if (err instanceof PoliciyError && err.name === 'SEARCH_POLICIY_ERROR') {
			return { loadError: err.cause as GraphQLError };
		} else if (err instanceof ZodError) {
			const { formErrors, fieldErrors } = err.flatten();
			return { formErrors, fieldErrors };
		} else if (isHttpError(err)) {
			throw error(err.status, err.body);
		} else if (isAppError(err)) {
			throw error(500, err);
		}

		log.error('policies:actions:load:error:', err);
		Sentry.setContext('source', { code: 'policy' });
		Sentry.captureException(err);
		throw error(500, getAppError(err));
	}
}) satisfies PageServerLoad;

export const actions = {
	delete: async ({ request, fetch }) => {
		const formData = await request.formData();
		const { id } = await zx.parseForm(formData, policiyDeleteSchema);

		const variables = { id: id + 'eee' };

		try {
			const data = await deletePolicyStore.mutate(variables, {
				metadata: { backendToken: 'token from tokenStore' },
				fetch
			});

			const actionResult = data.delete_tz_policies_by_pk;
			if (!actionResult) throw new NotFoundError('policy not found');

			return {
				actionResult
			};
		} catch (err) {
			console.debug('policies:actions:delete:error:', err);

			if (err instanceof NotFoundError && err.name === 'NOT_FOUND_ERROR') {
				return fail(404, { actionError: err });
			} else if (err instanceof ZodError) {
				// const { formErrors, fieldErrors } = err.flatten();
				return fail(400, { actionError: err });
			} else if (Array.isArray(err)) {
				return fail(400, { actionError: err[0] as GraphQLError });
			} else if (isHttpError(err)) {
				throw error(err.status, err.body);
			} else if (isAppError(err)) {
				throw error(500, err);
			}

			console.error('policies:actions:delete:error:', err);
			Sentry.setContext('source', { code: 'policy.delete' });
			Sentry.captureException(err);
			throw error(500, getAppError(err));
		}
	}
} satisfies Actions;

```

### Criticality

None

### LINKED ISSUE #832: Support for operationName request field
### Describe the feature

Would like to be able to pass `operationName` to our server. While this field isn't required in a standard setup, it's great for logging & tracing requests on the backend.

Doesn't seem to be a way currently to get the name of the operation within the fetch handler as only the raw text and hash are passed down.

Quick look seems it would be as simple as passing the artifact name down to the fetch handler here:

https://github.com/HoudiniGraphql/houdini/blob/30a754a8ec5d2ddde9ab54a104e6358f1f9c973d/packages/houdini/src/runtime/lib/network.ts#L297-L301

Won't be able to open a PR until the weekend, so opening this issue until then in case someone wants to take it.

### Criticality

nice to have

### LINKED ISSUE #840: Error: Could not find referenced fragment definition: FragmentName
### Describe the bug

Sometimes existing fragments don't seem to be processed early enough. When I `run build` twice, it often  – but not reliably – works. It seems to only happen when there are a couple more fragments

@AlecAivazis: I updated the reproduction from https://github.com/HoudiniGraphql/houdini/issues/819 with a bunch of other queries and fragments from my project. Maybe you can reproduce it like this? A slug you could use for testing is: `etcetera`

### Severity

blocking all usage of Houdini

just try to build it

### Reproduction

_No response_

### LINKED ISSUE #844: client not getting session data for mutate call
### Describe the bug

for **fetch** call, I am getting _session_ data as expected

image

But, for **mutate** call,  same client not printing  the _session_ data

image

My [client](https://github.com/xmlking/svelte-starter-kit/blob/main/src/lib/graphql/client.ts#L14-L19)
[hook](https://github.com/xmlking/svelte-starter-kit/blob/main/src/lib/server/middleware/houdini.ts#L18) where I set token into session. 

### Severity

https://github.com/xmlking/svelte-starter-kit

### Reproduction

https://xmlking@github.com/xmlking/svelte-starter-kit.git

### LINKED ISSUE #854: [ERROR] Could not resolve "HOUDINI_CONFIG_PATH"
### Describe the bug

Suddenly getting this error: 

```✘ [ERROR] Could not resolve "HOUDINI_CONFIG_PATH"

    node_modules/.pnpm/houdini@0.20.3/node_modules/houdini/build/lib-esm/index.js:61460:43:
      61460 │   return defaultConfigValues((await import("HOUDINI_CONFIG_PATH")).default);
            ╵                                            ~~~~~~~~~~~~~~~~~~~~~

  You can mark the path "HOUDINI_CONFIG_PATH" as external to exclude it from the bundle, which will
  remove this error. You can also add ".catch()" here to handle this failure at run-time instead of
  bundle-time.

10:53:26 AM [vite] error while updating dependencies:
Error: Build failed with 1 error:
node_modules/.pnpm/houdini@0.20.3/node_modules/houdini/build/lib-esm/index.js:61460:43: ERROR: Could not resolve "HOUDINI_CONFIG_PATH"
    at failureErrorWithLog (/home/dexter/Projects/wolbodo/members/node_modules/.pnpm/esbuild@0.16.17/node_modules/esbuild/lib/main.js:1604:15)
    at /home/dexter/Projects/wolbodo/members/node_modules/.pnpm/esbuild@0.16.17/node_modules/esbuild/lib/main.js:1056:28
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

Unsure what triggered it and how to fix it.

### Severity

blocking all usage of Houdini

The code is available on https://github.com/wolbodo/members/tree/svelte-kit%40v1

Its being triggered just by setting up the app and running `pnpm run dev`

### Reproduction

_No response_

### LINKED ISSUE #856: TypeError: Class extends value undefined is not a constructor or null
### Describe the bug

I am using `houdini-plugin-svelte-global-stores` , and when I test this example project, getting the following error
```
TypeError: Class extends value undefined is not a constructor or null
    at /$houdini/plugins/houdini-svelte/stores/listPokemon.js:4:39
    at async instantiateModule (file:///Users/schintha/Developer/Work/SPA/bbb/houdini-issues/node_modules/.pnpm/vite@4.0.4/node_modules/vite/dist/node/chunks/dep-5e7f419b.js:52224:9)

```

### Severity

annoyance

https://github.com/xmlking/houdini-issues
pnpm i
pnpm dev
open http://localhost:5173/test1

### Reproduction

https://github.com/xmlking/houdini-issues

### LINKED ISSUE #884: The type of `Props` is not carried over correctly from generated variables function type for component queries
### Describe the bug

Typescript complains it does not known the property even though it is present in the tooltip:

image

The generated types seem to be correct:
image

Also the return type of the function is generic (or rather not even found) although the correct type is imported (but unused).

### Severity

annoyance

Have this component:

```svelte
<script lang="ts">
  import { graphql } from '$houdini';
  import type { GetCatalogsAdminExpansionVariables } from './$houdini';

  export let parentCatalogId: number;

  export const _GetCatalogsAdminExpansionVariables: GetCatalogsAdminExpansionVariables = ({
    props,
  }) => ({


    parentCatalogId: props.parentCatalogId,
  });

  $: store = graphql(`
    query GetCatalogsAdminExpansion($parentCatalogId: Int!) {
      catalogs(parentCatalogId: $parentCatalogId) {
        id
      }
    }
  `);
</script>
```

with this generated `$houdini.ts`:
```ts

import type { GetCatalogsAdminExpansion$input } from '../../../../../artifacts/GetCatalogsAdminExpansion'
                    
import CatalogCardExpansion from './CatalogCardExpansion.svelte'
		import type { ComponentProps } from 'svelte'
type CatalogCardExpansionProps = ComponentProps<CatalogCardExpansion>

export type GetCatalogsAdminExpansionVariables = <_Props = CatalogCardExpansionProps>(args: { props: _Props }) => FragmentQueryVars$input
        
```

Notice typescript is complaining about `props.parentCatalogId`.
Also the return type of the function is `FragmentQueryVars$input` but should probably be `GetCatalogsAdminExpansion$input`.

### Reproduction

_No response_

### LINKED ISSUE #886: v8 error: JavaScript heap out of memory when generating types for generic types with inline fragments
### Describe the bug

In our project at work today, I wrote a query that looks somewhat similar to this:
```gql
query MyAnimalQuery {
  animal(id: "some_cat") {
    id

    ...AnimalDetails

    ... on Cat {
      ...CatThings
    }
  }
}

fragment AnimalDetails on Animal {
  id
  name
  description

  # Need this so that Houdini can generate the __typename properly
  ... on Cat {
    id
  }
}

fragment CatThings on Cat {
  id
  name
  nrOfGlassesKnockedOver

  ...AnimalDetails
}
```
Each fragment is in a separate component, where they're used to display the proper data in the UI.

Upon running `houdini generate`, it hangs for a while, eats up 4GB of RAM, and then v8 crashes because it ran out of heap memory:
```
<--- Last few GCs --->

[7752:0000021720CE3E20]    33580 ms: Mark-sweep 4054.5 (4128.4) -> 4044.1 (4134.1) MB, 2030.9 / 0.0 ms  (average mu = 0.195, current mu = 0.073) allocation failure scavenge might not succeed
[7752:0000021720CE3E20]    35686 ms: Mark-sweep 4059.9 (4134.1) -> 4047.5 (4137.4) MB, 2094.9 / 0.0 ms  (average mu = 0.107, current mu = 0.005) allocation failure scavenge might not succeed


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 00007FF73BEF0AAF v8::internal::CodeObjectRegistry::~CodeObjectRegistry+124015
 2: 00007FF73BE7C866 v8::internal::wasm::WasmCode::safepoint_table_offset+64182
 3: 00007FF73BE7D8E2 v8::internal::wasm::WasmCode::safepoint_table_offset+68402
 4: 00007FF73C7B1CE4 v8::Isolate::ReportExternalAllocationLimitReached+116
 5: 00007FF73C79C2AD v8::SharedArrayBuffer::Externalize+781
 6: 00007FF73C63F88C v8::internal::Heap::EphemeronKeyWriteBarrierFromCode+1468
 7: 00007FF73C64C539 v8::internal::Heap::PublishPendingAllocations+1129
 8: 00007FF73C64950A v8::internal::Heap::PageFlagsAreConsistent+2842
 9: 00007FF73C63C169 v8::internal::Heap::CollectGarbage+2137
10: 00007FF73C63A320 v8::internal::Heap::AllocateExternalBackingStore+2000
11: 00007FF73C65EEA6 v8::internal::Factory::NewFillerObject+214
12: 00007FF73C391645 v8::internal::DateCache::Weekday+1797
13: 00007FF73C83F9C1 v8::internal::SetupIsolateDelegate::SetupHeap+494417
14: 0000021722D4A7A9
```

Commenting out the `... on Cat` in the `AnimalDetails` fragment fixes it.
Funnily enough, if I change it to `... on Dog`, houdini generate works just fine.

This might be because of me slightly mis-using the powers of GraphQL, but the query itself works just fine on the GraphQL server, so I thought I'd report it.
In the end I worked around this by fetching the data in a cleaner way, so it's not a blocker atm.

### Severity

1. Clone the linked repo, on the `bug-repro/js-out-of-memory` branch (example code is in the e2e/sveltekit/src/routes/js-out-of-memory directory)
2. Run `npx houdini generate` in the e2e/sveltekit project
3. Wait and see the error pop up after a while (also note the RAM usage in task manager)
4. Comment out the `... on Cat` in the `AnimalDetails` fragment, and see houdini generate completing successfully

### Reproduction

https://github.com/SeppahBaws/houdini/tree/bug-repro/js-out-of-memory/e2e/sveltekit/src/routes/js-out-of-memory

### LINKED ISSUE #888: pagination loadNextPage not passing auth headers to backend
### For "Load More" button,   it is not passing headers. i.e.. `fetchParams` in client is not getting invoked.

 
Screen Shot 2023-02-02 at 3 33 13 PM

### For "Refresh" button, it is passing headers as coded in the **client**
Screen Shot 2023-02-02 at 3 24 13 PM

### Severity

blocking an upgrade

https://github.com/xmlking/svelte-starter-kit/blob/main/src/routes/(play)/play/houdini/%2Bpage.svelte

### Reproduction

https://github.com/xmlking/svelte-starter-kit/blob/main/src/routes/(play)/play/houdini/%2Bpage.svelte

### LINKED ISSUE #905: @include and @skip directives are stripped out of the query
### Describe the bug

(houdini version: `1.0.0-next.19`)
When I execute this query, the `@skip` and `@include` directives are stripped out, and an invalid query is formed.

The query in my code:
```gql
query TestQuery($userId: ID!, $scopeOnUser: Boolean!) {
  users @skip(if: $scopeOnUser) {
    edges {
      node {
        id
        name
      }
    }
  }
  user(id: $userId) @include(if: $scopeOnUser) {
    id
    name
  }
}
```
The query that is being sent to the graphql endpoint:
```gql
query TestQuery($userId: ID!) {
  users {
    edges {
      node {
        id
        name
      }
    }
  }
  user(id: $userId) {
    id
    name
  }
}
```

### Severity

1. clone my repro fork
2. run the project from the `bug-repro/include-directive-stripped-out` branch
3. open the `http://localhost:5173/include-directive-stripped-out` route
4. click the "Load data" button
5. inspect the query sent to the graphql backend
6. see the `@include` and `@skip` directives no longer being there, and making the query invalid because of the unset "userId" variable

### Reproduction

https://github.com/SeppahBaws/houdini/tree/bug-repro/include-directive-stripped-out

### LINKED ISSUE #916: LoadInput Type generation incorrect
### Describe the bug

If there is both a page and a layout query, the type generation creates 2 separate types for the LoadInput value.

This is due to the houdini-svelte codegen plugin. 

https://github.com/HoudiniGraphql/houdini/blob/f535af8de5838722888de944ca4a21d7450f0b8e/packages/houdini-svelte/src/plugin/codegen/routes/index.ts#L185

### Severity

annoyance

1. Initialize a project
2. Have both a layout load and page load queries with inputs
3. Generated see generated load input incorrect

### Reproduction

_No response_

AlecAivazis:
This PR merges `next` into `main` and kicks off the release for `1.0` 🎉 

If you're reading this, please find an application to test it against  and report back 🙌 You can find the latest version as `@next` and the guide for upgrading can be found here: http://next.houdinigraphql.com/guides/release-notes

closes #905
closes #365
closes #832 
closes #670
fixes #844
closes #287
fixes #856
closes #821
closes #466
closes #383 
closes #884 
closes #886
closes #840
closes #888 
closes #916


fehnomenal:
It's nice to work again on my project using houdini. But, oh boy... I need to upgrade from sveltekit 1.0.0-next.398 and houdini 0.15.6.

Anyway, I noticed in the release notes `subscriptions` is mentioned but the correct import (at least as of 1.0.0-next.8) is `subscriptionPlugin`.

AlecAivazis:
Nice to hear from you again! 👋  I just pushed up a fix for the release notes.  I'm still going back and forth on that plugin name. it's basically the last thing left on my list😅 

> sveltekit 1.0.0-next.398 and houdini 0.15.6.

wow that's gonna be a journey! That's the old routing stuff for kit right?  I just checked and that's even before we moved everything to the vite plugin! Well I hope you enjoy all of the changes we've been making. At the very least you should enjoy not having to run `generate` ever again. 

As always, let me know if you run into any issues with the release guides 👍

fehnomenal:
> wow that's gonna be a journey! That's the old routing stuff for kit right?

Yeah, it's really nasty. But hey, just a few hours in and some single pages already load without errors :joy:

I really love all the changes in houdini and the migration went very smooth so far :tada:

rmarscher:
@AlecAivazis The only issue I've seen so far... testing against 1.0.0-next.14, I have some component queries on pages that don't have any other client-side load and are hitting the `"client hasn't been initialized"` exception in the generated runtime client.js.

I define the component query in an external .gql file, import the generated store and construct it in the component and then on a button click it calls `await store.fetch(...)`.  

Essentially:

```svelte
<script lang="ts">
    import { MyQueryStore } from '$houdini'

    const store = new MyQueryStore()

    function onClick() {
        const result = await store.fetch()
        console.log('do something with', result)
    }
</script>

<button on:click="{onClick}" type="button">Click me</button>
```


I added this in our top level layout file and it fixes the issue.

```ts
import { initClient } from '$houdini/plugins/houdini-svelte/runtime/client';

// ...

onMount(async () => {
    initClient();
    // ...
}
```

AlecAivazis:
Oh thanks for reporting that! `fetch` and `subscribe` shuold both be making sure its called but maybe i missed some edge case - i'll take a look. The idea was that you should have to call that ever

AlecAivazis:
@rmarscher - should be fixed in `v1.0.0-next.17` 👍

rmarscher:
> @rmarscher - should be fixed in `v1.0.0-next.18` 👍

Great. We're testing against v1.0.0-next.19. Confirmed it's working without needing to call initClient ourselves now. Thanks!

PR #874, opened 2023-02-01
A couple typos in new client plugins doc

rmarscher:
I noticed these while following the 1.0 migration docs and migrating our custom client fetch to a network plugin. Thanks!

### DIFF
--- a/site/src/routes/api/client-plugins/+page.svx
+++ b/site/src/routes/api/client-plugins/+page.svx
@@ -153,7 +153,7 @@ const simpleFetchPlugin: ClientPlugin = () => {
     return {
         async network(ctx, { resolve, next }) {
             const result = await fetch('...', {
-                body: JSON.strinify({ query: ctx.text })
+                body: JSON.stringify({ query: ctx.text })
             })
 
             // in reality we need to pass more here
@@ -322,7 +322,7 @@ to your hooks. For a full summary, please refer to the the `ClientPluginEnterHan
 ## Type Definitions
 
 The best source of truth for the type definitions are exported from your `$houdini`
-package. You can see them [here](https://github.com/HoudiniGraphql/houdini/tree/main/packages/houdini/src/runtime/client/DocumentStore.ts). They've been summarized below for reference but this copy may be out of date.
+package. You can see them [here](https://github.com/HoudiniGraphql/houdini/blob/main/packages/houdini/src/runtime/client/documentStore.ts). They've been summarized below for reference but this copy may be out of date.
 If you find a discrepancy, please let us know on GitHub.
 
 ```typescript


AlecAivazis:
oh nice! thanks for fixing those 👍

AlecAivazis (approved):

AlecAivazis:
Im curious to hear more about your experience with the client plugins! Could you give me some more info? As far as I know, you're the first to write one that isn't me so any feedback on docs, mental model, etc would be huge 🙌

AlecAivazis:
Im curious to hear more about your experience with the client plugins! Could you give me some more info? As far as I know, you're the first to write one that isn't me so any feedback on docs, mental model, etc would be huge 🙌

rmarscher:
I love the client plugins. The model makes sense to me. I like the composability options and it was pretty simple to port what we were doing pre-1.0 to a client plugin. We're using it for two things now:

1. Do some additional data marshaling between our typescript objects and our graphql schema - we have some generated fields and some fields that we automatically encode/decode to JSON to workaround graphql's inability to store data with arbitrary keys.
2. In server-side code that doesn't have a current user session -- like webhook handlers -- we prepare a custom fetch function that signs requests with IAM credentials (we use AWS AppSync as our graphql server and it supports IAM auth via these signed requests). We have a special IAM user that we treat as our "system" user for accessing the GraphGL API. We pass that custom fetch function down through the houdini metadata and in our network plugin, if it sees metadata.fetch is set, then it calls that instead of the default ctx.fetch.

Here's the network plugin:

```ts
async network(ctx, { resolve }) {
	let result: Response;
	// Ability for server-side queries to pass an alternate fetch
	if (ctx.metadata?.['fetch']) {
		const _fetch = ctx.metadata['fetch'] as typeof fetch;
		result = await _fetch(ctx.text, ctx.variables);
	} else {
		const options: RequestInit = {
			method: 'POST',
			credentials: 'include',
			headers: {
			},
			body: JSON.stringify({
				query: ctx.text,
				variables: ctx.variables,
			}),
		};
		result = await ctx.fetch('/graphql', options);
	}
	const clonedResult = result.clone();
	try {
		resolve(ctx, await result.json());
	} catch (e) {
		console.log('Problem parsing result json: ', await clonedResult.text());
		throw e;
	}
},
```

And here's the custom fetch which we import in *.server.ts files and pass in our houdini fetch and mutate calls.

```
import crypto from '@aws-crypto/sha256-js';
import { SignatureV4 } from '@aws-sdk/signature-v4';
import type { HttpMessage, EndpointV2 } from '@aws-sdk/types';
import { HttpRequest } from '@aws-sdk/protocol-http';
import {
	APPSYNC_AWS_REGION,
	APPSYNC_AWS_ACCESS_KEY_ID,
	APPSYNC_AWS_SECRET_ACCESS_KEY,
} from '$env/static/private';

const { Sha256 } = crypto;
const GRAPHQL_ENDPOINT = import.meta.env.GRAPHQL_URL;
const AWS_REGION = APPSYNC_AWS_REGION || 'us-east-1';

const endpoint = new URL(GRAPHQL_ENDPOINT);

type HttpRequestOptions = Partial<HttpMessage> &
	Partial<EndpointV2> & {
		method?: string;
	};

const signer = new SignatureV4({
	credentials: {
		accessKeyId: APPSYNC_AWS_ACCESS_KEY_ID,
		secretAccessKey: APPSYNC_AWS_SECRET_ACCESS_KEY,
	},
	region: AWS_REGION,
	service: 'appsync',
	sha256: Sha256,
});

export async function serverGraphqlFetch(query: string, variables: { [k: string]: unknown }) {
	const options: HttpRequestOptions = {
		method: 'POST',
		hostname: endpoint.host,
		path: endpoint.pathname,
		headers: {
			Host: endpoint.host,
		},
		body: JSON.stringify({
			query,
			variables,
		}),
	};

	const requestToBeSigned = new HttpRequest(options);
	const signed = await signer.sign(requestToBeSigned);
	const gqlRequest = new Request(endpoint, signed);
	return fetch(gqlRequest);
}
```

rmarscher:
Sorry that wasn't really any useful feedback. Just praise :-)

AlecAivazis:
Oh don't worry - praise is fine too 😅 

I'm super happy to hear you had an easy time adapting to the new world. I was mostly concerned about the overall complexity of having multiple phases, rules, terms, etc. but it sounds like you were able to navigate it well. Thanks for the feedback!

PR #887, opened 2023-02-02
Fix typings of generated variable function for component queries

### LINKED ISSUE #884: The type of `Props` is not carried over correctly from generated variables function type for component queries

fehnomenal:
Fixes #884

### DIFF
--- a/.changeset/ten-plums-fetch.md
+++ b/.changeset/ten-plums-fetch.md
@@ -1,0 +1,5 @@
+---
+'houdini-svelte': patch
+---
+
+Fix typings of generated variable function for component queries
--- a/packages/houdini-svelte/src/plugin/codegen/components/index.ts
+++ b/packages/houdini-svelte/src/plugin/codegen/components/index.ts
@@ -131,7 +131,7 @@ import type { ${query.name}$input } from '${path
 				`
 export type ${config.variableFunctionName(
 					query.name
-				)} = <_Props = ${prop_type}>(args: { props: _Props }) => FragmentQueryVars$input
+				)} = <_Props extends ${prop_type}>(args: { props: _Props }) => FragmentQueryVars$input
         `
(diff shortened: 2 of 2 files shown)


AlecAivazis:
Thanks for the quick fix!

PR #904, opened 2023-02-06
Random tweaks

### LINKED ISSUE #905: @include and @skip directives are stripped out of the query

AlecAivazis:
Fixes #905

This PR makes some minor adjustments to make plugin authoring a little more streamlined and some other tweaks 
- export `$options` type for enum values
- `pluginData` always has _a_ value for a plugin that defines the key
- Houdini's enum values are now strings so that users don't accidentally import `houdini` into their runtime
- plugins must define their custom config in a separate file (so we can ensure its safe to bundle)


site/src/routes/guides/release-notes/+page.svx
```diff
@@ -370,6 +370,37 @@ in m ind that global query stores are dangerous when used with SSR:
...
+
+function onClick() {
+	// could use a utility object
+	update.mutate({ variables: { input: MyEnum.Value1 } })
+	// or pass a string directly
+	update.mutate({ variables: { input: 'Value1' } })
+}
+```
+
+A side effect of this change is that yif ou need to refer to the list of all values for
```
SeppahBaws, line 389: There's a small typo in here 🙂 
```suggestion
A side effect of this change is that if you need to refer to the list of all values for
```

AlecAivazis: Whoops! Nice catch 👍

Issue #926, opened 2023-02-12
Cannot fetch schema with cookie authentication


yaroslavros:
### Describe the bug

Cookie authentication always fails when fetching schema because cookie header value contains "=" and it is getting lost due to `header.split("=")`

Replacing `header.split("=")` with `header.split(/=(.*)/s)` in cmd/pullSchema.ts and cmd/init.ts fixes the issue

### Severity

serious, but I can work around it

### Steps to Reproduce the Bug

Attempt to use init or pull-schema against graphql API that uses Cookie authentication or any other header containing "="

AlecAivazis:
hey @yaroslavros! Thanks for finding the fix - mind submitting a PR?

AlecAivazis:
Thanks again for submitting the fix!

PR #929, opened 2023-02-12
HTTP headers values should allow = sign for Cookie auth

yaroslavros:
HTTP headers with equal sign (=) in the value are now correctly passed to init and pull-schema command line params. This allows cookie-based authentication to graphql among other things.

Fixes [#926](https://github.com/HoudiniGraphql/houdini/issues/926)


AlecAivazis (approved):
Thanks!!

PR #931, opened 2023-02-12
Fix header key in API -> Client docs

destructobeam:
Fixes N/A


AlecAivazis (approved):
Thanks!

PR #939, opened 2023-02-15
Svelte: Warn if "include" config key doesn't specify js files

luhagel:

This PR is the result of a thread @AlecAivazis and I had on discord, where all the generated code was simply *missing* / not there in the first case, which was very difficult to debug on my own.

Now, we throw a warning during the `houdini-svelte` codegen step.

### DIFF
--- a/packages/houdini-svelte/src/plugin/codegen/routes/index.ts
+++ b/packages/houdini-svelte/src/plugin/codegen/routes/index.ts
@@ -3,7 +3,7 @@ import type { Config, GenerateHookInput } from 'houdini'
 import { fs, path } from 'houdini'
 
 import type { HoudiniSvelteConfig } from '../..'
-import type { Framework } from '../../kit'
+import { Framework, plugin_config } from '../../kit'
 import { type_route_dir, stores_directory_name, store_suffix, walk_routes } from '../../kit'
 import { houdini_afterLoad_fn, houdini_before_load_fn, houdini_on_error_fn } from '../../naming'
 import { route_params } from '../../routing'
@@ -56,7 +56,7 @@ export default async function svelteKitGenerator(
 				// Include path doesn't include js
 				!config.include.some((path) => path.includes('js')) &&
 				// And the plugin isn't configured to run in static mode
-				!(config.pluginConfig('houdini-svelte') as HoudiniSvelteConfig).static
+				!plugin_config(config).static
 			) {
 				console.error(
 					`⚠️ You are using at least one page/layout query but aren't "include"ing .js files in your config. \n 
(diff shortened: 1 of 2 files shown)


packages/houdini-svelte/src/plugin/codegen/routes/index.ts
```diff
@@ -49,6 +50,20 @@ export default async function svelteKitGenerator(
 				}
 			}
 
+			if (
+				// User uses layout/page queries
+				(uniquePageQueries.length > 0 || uniqueLayoutQueries.length > 0) &&
+				// Include path doesn't include js
+				!config.include.some((path) => path.includes('js')) &&
+				// And the plugin isn't configured to run in static mode
+				!(config.pluginConfig('houdini-svelte') as HoudiniSvelteConfig).static
```
AlecAivazis, line 59: Mind tweaking this to use the `plugin_config` utility from [here](https://github.com/HoudiniGraphql/houdini/blob/main/packages/houdini-svelte/src/plugin/kit.ts#L361)? It's really hidden so I don't blame you for not seeing it existed but it will clean this up a bit and apply any defaults

AlecAivazis (changes requested):

packages/houdini-svelte/src/plugin/codegen/routes/index.ts
luhagel: Done

packages/houdini-svelte/src/plugin/codegen/routes/index.ts
```diff
@@ -49,6 +55,20 @@ export default async function svelteKitGenerator(
 				}
 			}
 
+			if (
+				// User uses layout/page queries
+				(uniquePageQueries.length > 0 || uniqueLayoutQueries.length > 0) &&
+				// Include path doesn't include js
+				!config.include.some((path) => path.includes('js')) &&
+				// And the plugin isn't configured to run in static mode
+				!plugin_config(config).static
```
AlecAivazis, line 64: So even if they don't use any generated load, they still need to have `.js` in their `include` to get session support. So I think something we want something like this:
```suggestion
				// Include path doesn't include js
				!config.include.some((path) => path.includes('js')) &&
				(
				  // And the plugin isn't configured to run in static mode
				  !plugin_config(config).static ||
				  // User uses layout/page queries
				  (uniquePageQueries.length > 0 || uniqueLayoutQueries.length > 0)
				)
```

AlecAivazis (changes requested):

AlecAivazis (approved):
Thanks again!

PR #950, opened 2023-02-17
@include and @skip now add "undefined" to the generated TypeScript type

### LINKED ISSUE #945: @include and @skip directives don't generate nullable TypeScript types
### Describe the bug

Upon generating typings for a query with `@include` and `@skip` directives in it, Houdini doesn't generate these types as nullable.

For this simple query:
```gql
query MyTestQuery($sayHello: Boolean!) {
  hello @include(if: $sayHello)
  user(id: "1", snapshot: "my-user") {
    id
    name
  }
}
```

Houdini generates this type:
```ts
export type MyTestQuery$result = {
    readonly hello: string;
    readonly user: {
        readonly id: string;
        readonly name: string;
    };
};
```

However, if `$sayHello` is set to false, the api will return an object without that field, and in the `$store.data` object, that field will contain `undefined`. But the types don't reflect this, which can lead to confusion. I would expect that the `hello` field in this case would look something like this:
```diff
export type MyTestQuery$result = {
-     readonly hello: string;
+     readonly hello: string | undefined;
    readonly user: {
```

My personal choice would be to use `string | undefined`, because that's what JavaScript will populate that object with by default. Making it `string | null` would lead to weird bugs with `!== null` checking, when the actual data would be `undefined`.

### Severity

annoyance

1. paste the example query in a test project
2. run `houdini generate`
3. observe the `hello` field to be not nullable
4. if you run this query, you will see that the `hello` field will contain `undefined`

### Reproduction

_No response_

SeppahBaws:
Fixes #945 

### DIFF
--- a/.changeset/nice-keys-add.md
+++ b/.changeset/nice-keys-add.md
@@ -1,0 +1,5 @@
+---
+'houdini': minor
+---
+
+@include and @skip now add "undefined" to the generated TypeScript type
--- a/packages/houdini/src/codegen/generators/typescript/inlineType.ts
+++ b/packages/houdini/src/codegen/generators/typescript/inlineType.ts
@@ -205,6 +205,19 @@ export function inlineType({
 					allOptional,
 				})
 
+				// check if we have an @include or @skip directive
+				const hasIncludeOrSkipDirective =
+					selection.directives &&
+					selection.directives.filter(
+						(directive) =>
+							directive.name.value === 'include' || directive.name.value === 'skip'
+					).length > 0
+
+				// if we do, add "undefined" to the type union
+				if (hasIncludeOrSkipDirective) {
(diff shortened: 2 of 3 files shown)


AlecAivazis (approved):
wow! tests, changeset, comments, and a nice and small diff. What more could I ask for?

.changeset/nice-keys-add.md
```diff
@@ -0,0 +1,5 @@
+---
+'houdini': minor
```
AlecAivazis, line 2: Doh! This should have been `patch` 🤦

SeppahBaws: Oh oops! First time doing something with changesets, so I'm not too familiar with it 😬

PR #952, opened 2023-02-20
Add failing spec for single page bidirectional cursor pagination

douglasward:
Fixes https://github.com/HoudiniGraphql/houdini/discussions/944

This PR adds a failing spec for "single page" bidirectional cursor pagination. If the default cache policy is used (`CacheOrNetwork`) then no network request is made and the shown page is not updated. If the policy is explicitly set to `NetworkOnly` then a network request is made and the shown page is updated correctly.

I also noticed that `last` and `before` / `first` and `after` must be explicitly set to `null` in order for the query to be fetched correctly. There is apparently no behind-the-scene-magic here.


AlecAivazis:
Amazing! Thanks so much for putting together the e2e test to reproduce. I'll try to find some time in the next few days to check it out 👍

jycouet:
`Vercel` is kinda annoying! (I'm probably doing something wrong... but what?!) :(

douglasward:
Thank you, @jycouet! That certainly helps unblock us for now - but would be great to to get "page by page" support baked in. 

I like your idea of being able to define the mode in the paginate directive, rather than adding a new directive. As a user I think I would look to configure the "paginate" directive / the current pagination setup to get want I want rather than looking for a different directive to use. 

(And I'm afraid I can't help regarding vercel ;) but hope you manage to figure it out!)

SeppahBaws:
In our current setup, we do page-by-page pagination manually, so having this baked-in would be very useful!

For the implementation, I think it would be best if there would be a default to set in the `houdini.config.js` (and if it isn't set, it would do infinite scroll by default), as well as being able to configure it on the query itself with the directive.
This way, (in our use case) we could configure Houdini to use page-by-page by default, and then we can simply specify a certain connection to use infinite scrolling instead.

just my two cents on this 🙂

jycouet:
Perfect comment @SeppahBaws, it's already prepared like that 😉
I should send to PR this week 👍

Thx @douglasward for opening the exchange & e2e tests 👍👍👍

PR #954, opened 2023-02-23
Value masking and fine-grain reactivity

AlecAivazis:
This PR implements some massive performance improvements to the runtime.  Previously, cache subscriptions were at the query level which meant that if one entry in a list of 10k changed, the whole list gets re-rendered. This PR changes that so that fragments are individually responsible for subscribing to the cache. This means that in the previous example, only the one fragment that handles the updated row would re-render.

Since fragments contain their own values now, the return value from `fragment` no longer hold transient fields from unrelated fragments. This has already been imposed by the generated types so I feel comfortable releasing this as a minor update since it's mostly an internal improvement on the documented behavior. 

As a side note, the diff on this PR is so nasty because it introduced a few new fields to the artifacts (`visible` and `fragments`) which forced a lot of snapshot updates.


e2e/sveltekit/src/routes/stores/connection-fragment/+page.svelte
```diff
@@ -0,0 +1,30 @@
...
+      user(id: "1", snapshot: "connection-fragment") {
+        friendsConnection(first: 2) {
+          ...ConnectionFragment
+        }
+      }
+    }
+  `);
+
+  $: frag = fragment(
+    $queryResult.data?.user?.friendsConnection ?? null,
```
AlecAivazis, line 15: @jycouet btw, the `?? null` is no longer necessary. This PR updated the typedefs to treat `undefined` as `null` for the fragment reference

jycouet: Sounds great 👍
On my app, it seems to have no issue.

PR #958, opened 2023-02-27
@Paginate two modes

jycouet:
Fixes #944

New feature: paginate directive now supports a PageByPage mode (in addition to the Infinite one already present)


packages/houdini-svelte/src/runtime/stores/pagination/offset.ts
```diff
@@ -61,14 +61,27 @@ export function offsetHandlers<_Data extends GraphQLObject, _Input extends {}>({
...
+				})
+			} else {
+				await parentFetchUpdate({
+					variables: queryVariables as _Input,
+					fetch,
+					metadata,
+					policy: CachePolicy.NetworkOnly,
+					session: await getSession(),
+				})
+			}
```
AlecAivazis, line 84: we still want the session passed here. maybe something like:

```suggestion
			const targetFetch = isPageByPage ? parentFetch :  parentFetchUpdate
			await targetFetch({
				variables: queryVariables as _Input,
				fetch,
				metadata,
				policy: isPageByPage ?  artifact.policy : CachePolicy.NetworkOnly,
				session: await getSession(), // Hum?
			})
```

packages/houdini/src/codegen/generators/definitions/schema.test.ts
```diff
@@ -28,6 +28,11 @@ test('adds internal documents to schema', async function () {
 			  NetworkOnly
 			}
 
+			enum PaginateMode {
```
AlecAivazis, line 31: What do you think of Window instead of "PageByPage"? My thinking is that Infinite is also loading page-by-page which makes me think we might want a different word 🤔

AlecAivazis (changes requested):
Nice job! Just a few small comments and it should be good to go 👍

packages/houdini/src/codegen/generators/definitions/schema.test.ts
jycouet: Humm, not a big fan of Windows, but I understand the concern.
What I had in mind was the way paginate is displaying the result, not how it's loading it.

AlecAivazis (approved):

PR #961, opened 2023-03-01
Pass ctx to the throwOnError error handler

SeppahBaws:
As discussed in Discord ([link to conversation](https://discord.com/channels/1024421016405016718/1077532514928820225)), passing the `ctx` to throwOnError in the client would be useful when trying to figure out which query exactly it is that is giving the error.

~~I've gone ahead and implemented a demo of this in the e2e project, but it should still work if you don't need to use the ctx.~~ It was messing up the e2e tests, so I have reverted this.

### DIFF
--- a/.changeset/blue-toes-carry.md
+++ b/.changeset/blue-toes-carry.md
@@ -1,0 +1,5 @@
+---
+'houdini': patch
+---
+
+Pass ctx to error handler in throwOnError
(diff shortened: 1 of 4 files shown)


SeppahBaws:
I noticed that I missed a bit in the docs under the "Type definitions" section, will fix in a bit

SeppahBaws:
Alright it should be all good now 😅

AlecAivazis (approved):
Nice work! Thanks again

PR #964, opened 2023-03-02
fix: svelte only

jycouet:
Fixes #959

Various tweaks in the init and framework detection


packages/houdini-svelte/src/plugin/index.ts
```diff
@@ -175,6 +176,11 @@ export const error = svelteKitError
 		// try to import the kit module
 		try {
 			await import('@sveltejs/kit')
+			// TODO => Find a good idea to detect if we are using SvelteKit or Svelte
+			//   1/ Seems that await import('@sveltejs/kit') is not enough (depending on package manager)
+			//   2/ I was hoping that vite plugins would help... knowing which framework is used, but didn't find a good thing
+			//   3/ reading the vite config file? To see plugins import?
+			//   4/ Bring back a flag in the config?
```
jycouet, line 183: This is the main thing to fix; let's talk about it soon 👍

jycouet:
Yeahhh, it's working :D
If we agree on this, we need to change the "required" tests for PRs in general.

site/src/routes/guides/setting-up-your-project/+page.svx
```diff
@@ -159,21 +159,27 @@ export default config
 ```
 
 If you are using vite, you should use the `houdini/vite` plugin even if you aren't using kit.
+You will also need to make sure that the `$houdini` alias resolves to the directory in the root of your project.
```
AlecAivazis, line 162: I like these additions a lot 👍

AlecAivazis (dismissed):

AlecAivazis (approved):

PR #987, opened 2023-03-12
Implement `@required` directive

### LINKED ISSUE #923: @required directive
### Describe the feature

We should add a `@required` directive, similar to relay's that enforces that the necessary values exist or else the parent's value is set to `null`. 

This post does a good job summarizing the directive and it's use: https://relay.dev/blog/2023/01/03/resilient-relay-apps/

### Criticality

tadeokondrak:
Fixes #923

### DIFF
--- a/packages/houdini/src/codegen/generators/artifacts/selection.ts
+++ b/packages/houdini/src/codegen/generators/artifacts/selection.ts
@@ -216,7 +216,7 @@ function prepareSelection({
 			)
 			if (requiredDirective) {
 				fieldObj.nullable = false
-				fieldObj.bubbleNullability = true
+				fieldObj.required = true
 			}
 
 			// is there an operation for this field
@@ -303,9 +303,7 @@ function prepareSelection({
 
 				// bubble nullability up
 				if (
-					Object.values(fieldObj.selection?.fields ?? {}).some(
-						(field) => field.bubbleNullability
-					)
+					Object.values(fieldObj.selection?.fields ?? {}).some((field) => field.required)
 				) {
 					fieldObj.nullable = true
 				}
--- a/packages/houdini/src/codegen/generators/artifacts/selection.ts
+++ b/packages/houdini/src/codegen/generators/artifacts/selection.ts
@@ -208,15 +208,14 @@ function prepareSelection({
 			}
 
 			if (nullable) {
-				fieldObj.serverNullable = true
-				fieldObj.clientNullable = true
+				fieldObj.nullable = true
 			}
 
 			const requiredDirective = field.directives?.find(
 				(directive) => directive.name.value == config.requiredDirective
 			)
 			if (requiredDirective) {
-				fieldObj.clientNullable = false
+				fieldObj.nullable = false
 				fieldObj.bubbleNullability = true
 			}
 
@@ -308,7 +307,7 @@ function prepareSelection({
 						(field) => field.bubbleNullability
 					)
 				) {
-					fieldObj.clientNullable = true
+					fieldObj.nullable = true
 				}
 			}
 
--- a/packages/houdini/src/codegen/generators/artifacts/selection.ts
+++ b/packages/houdini/src/codegen/generators/artifacts/selection.ts
@@ -323,6 +323,13 @@ function prepareSelection({
 			// if we are looking at an interface
 			if (graphql.isInterfaceType(fieldType) || graphql.isUnionType(fieldType)) {
 				fieldObj.abstract = true
+				if (
+					Object.values(fieldObj.selection?.abstractFields?.fields ?? {}).some((fields) =>
+						Object.values(fields ?? {}).some((field) => field.required)
+					)
+				) {
+					fieldObj.abstractHasRequired = true
+				}
 			}
 
 			// if there is an existing value, merge them
--- a/packages/houdini/src/codegen/generators/artifacts/selection.ts
+++ b/packages/houdini/src/codegen/generators/artifacts/selection.ts
@@ -208,7 +208,16 @@ function prepareSelection({
 			}
 
 			if (nullable) {
-				fieldObj.nullable = true
+				fieldObj.serverNullable = true
+				fieldObj.clientNullable = true
+			}
+
+			const requiredDirective = field.directives?.find(
+				(directive) => directive.name.value == config.requiredDirective
+			)
+			if (requiredDirective) {
+				fieldObj.clientNullable = false
+				fieldObj.bubbleNullability = true
 			}
 
 			// is there an operation for this field
@@ -292,6 +301,15 @@ function prepareSelection({
 					typeMap,
 					abstractTypes,
 				})
+
+				// bubble nullability up
+				if (
+					Object.values(fieldObj.selection?.fields ?? {}).some(
+						(field) => field.bubbleNullability
+					)
+				) {
+					fieldObj.clientNullable = true
+				}
 			}
 
 			// any arguments on the list field can act as a filter
--- a/packages/houdini/src/codegen/generators/artifacts/selection.ts
+++ b/packages/houdini/src/codegen/generators/artifacts/selection.ts
@@ -303,7 +303,7 @@ function prepareSelection({
 
 				// bubble nullability up
 				if (
-					Object.values(fieldObj.selection?.fields ?? {}).some((field) => field.required)
+					Object.values(fieldObj.selection.fields ?? {}).some((field) => field.required)
 				) {
 					fieldObj.nullable = true
 				}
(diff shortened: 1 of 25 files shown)


packages/houdini/src/codegen/generators/artifacts/tests/artifacts.test.ts
```diff
@@ -5617,3 +5629,194 @@ test('fragment nested in root', async function () {
...
+
+		                "selection": {
+		                    "abstractFields": {
+		                        "fields": {
+		                            "Legend": {
+		                                "name": {
+		                                    "type": "String",
+		                                    "keyRaw": "name",
+		                                    "serverNullable": true,
+		                                    "clientNullable": false,
```
AlecAivazis, line 5716: Can you explain the difference between these 2 nullable flags?

AlecAivazis:
Awesome! Thanks for putting this together 👍

packages/houdini/src/codegen/generators/artifacts/tests/artifacts.test.ts
tadeokondrak: `serverNullable` is set for things that are nullable in the schema. `clientNullable` is always true when `serverNullable` is, but can be false when `@required` is used.

It might be possible to get rid of the distinction but the one piece of code that I left using `serverNullable` (the old value of `nullable`) is here: <https://github.com/tadeokondrak/houdini/blob/d6ceadae172fa0d02b682d48d0141dac6ab67e51/packages/houdini/src/runtime/cache/cache.ts#L321>. I don't understand what it's used for there.

(There's also `bubbleNullability`, which I only use during compile-time. It keeps the code simpler but I might be able to get rid of it and change the code instead.)

AlecAivazis:
Hopefully that kicks off the tests

packages/houdini/src/codegen/generators/artifacts/selection.ts
```diff
@@ -275,6 +283,13 @@ function prepareSelection({
 					typeMap,
 					abstractTypes,
 				})
+
+				// bubble nullability up
+				if (
+					Object.values(fieldObj.selection?.fields ?? {}).some((field) => field.required)
```
AlecAivazis, line 289: we need to be careful here about the selection since there could be abstract selections to consider. I think it would be best to avoid think about child fields here and instance handle the bubbling when we read values

AlecAivazis (changes requested):

packages/houdini/src/codegen/generators/artifacts/selection.ts
tadeokondrak: Hm, I don't understand. What are abstract selections in this context?

AlecAivazis: So right now, the generated artifact has 2 different places where selections are stored: the `fields` object that you found as well as `abstractFields` (another key on that `selection` object). Consider this query:

```
query { 
    node(id: "1") {
        ... on User { 
            parent { 
                name
            }
        }
        
        ... on Animal { 
            parent { 
                name
            }
        }
    }
}
```

Notice that the selection that we have to process depends on the actual type of the object at runtime? That's the reason for the `abstractFields` object: it gives us a type dependent place to hold onto a selection. As you can imagine, there are already a few edge-cases to consider here so I think rather than pushing yet-another-one onto the pile I'd like to see if we can find a way to implement this that keeps the selection logic only local (ie not concerned with other selections in the artifact).

If you're curious, we have to track each selection as a separate object because the actual types of the transient fields can be different (it only matters that the leaf nodes are the same). That is to say, that `parent` could be a different type in each branch.  Otherwise we could just encode `parentType` information

packages/houdini/src/codegen/generators/typescript/inlineType.ts
```diff
@@ -23,6 +23,7 @@ export function inlineType({
 	missingScalars,
 	includeFragments,
 	allOptional,
+	forceNonNull,
```
tadeokondrak, line 26: The typescript code still needs to handle the case described here: https://relay.dev/docs/next/guides/required-directive/#why-are-required-fields-in-an-inline-fragment-still-nullable

tadeokondrak: Done

packages/houdini/src/codegen/generators/artifacts/selection.ts
tadeokondrak: To match the Relay behaviour this code is still here, but the code now supports abstract fields.

packages/houdini/src/codegen/generators/typescript/typescript.test.ts
```diff
@@ -1556,6 +1557,330 @@ describe('typescript', function () {
 		`)
 	})
 
+	test('@required directive removes nullability from field', async function () {
```
AlecAivazis, line 1560: Thanks for the super thorough tests!!

packages/houdini/src/runtime/cache/cache.ts
```diff
@@ -951,10 +943,16 @@ class CacheInternal {
 				}
 			}
 
-			// regardless of how the field was processed, if we got a null value assigned
-			// and the field is not nullable, we need to cascade up
 			if (target[attributeName] === null && !nullable && !embeddedCursor) {
-				cascadeNull = true
+				// if we got a null value assigned and the field is not nullable, we need to cascade up
+				// except when it's an abstract type with @required children - then we return a dummy object
```
AlecAivazis, line 948: I think im missing something here. Can you explain the situation this is supposed to cover? I'm having a hard time seeing why we can't just enforce that `nullable` is `true` in when we have an abstract type. I'm also a bit confused why abstract types behave differently than normal types in general.

tadeokondrak: So I'm looking into this and it's a bit difficult to reason about this given that it seems that Houdini is already generating incorrect types for cases like the following: (schema lightly modified from e2e/kit)

```graphql
scalar DateTime

type User implements Node {
  id: ID!
  name: String!
  birthDate: DateTime
}

union UserResult = User | Message1

type Message1 {
	message: String!
}

type Query {
  userResult(id: ID!, forceMessage: Boolean!): UserResult!
}
```

```graphql
query ExampleQuery($id: ID!, $forceMessage: Boolean!) {
  userResult(id: $id, forceMessage: $forceMessage) {
    ... on User {
      name
      birthDate
    }
  }
}
```

We get:

```typescript
export type ExampleQuery$result = {
  readonly userResult: {
    readonly __typename: "User";
    readonly name: string;
    readonly birthDate: Date | null;
  };
};
```

This is wrong - if we then use:

```svelte
{#if $GQL_ExampleQuery.data}
  {$GQL_ExampleQuery.data.userResult.birthDate}
{/if}
```

We can get `undefined` displayed if `userResult` was actually a `Message1`. TypeScript thinks that the type of `birthDate` is `Date`, not `Date | undefined`.

The way that we deal with this problem will decide how we should handle similar cases with `@required`.

AlecAivazis: Oh nice catch, I think all we need to do is add `{}` to the generated type for any abstract selection set. Typescript seems to behave correctly: https://www.typescriptlang.org/play?#code/C4TwDgpgBAYg9nKBeKBvKBDAXAZ2AJwEsA7AcygF8oAfNCgKHoGM5i8oAzBZOzHWBIy5wAdBiA

tadeokondrak: Great, I'll do that. Coming back to the original topic of this thread, this code means we can do the same thing (add `{}` as a variant) and not have it be null when we're missing a required field. I could give a more detailed example for that case too.

As far as I understand right now, the code this thread is about is still necessary, but we can remove the `__typename` constant and simply return `{}` when a required field is missing.

tadeokondrak: I've pushed a commit that implements adding `{}` as an option to *non-exhaustive* abstract selections - meaning if you enumerate all possible types of an interface or union you don't need to handle the case of it being none of them. Let me know if you'd prefer the simpler case of always requiring the user to handle `{}`.

But here's an example showing that we can't just use `{}` and we need a dummy __typename like Relay: https://www.typescriptlang.org/play?#code/C4TwDgpgBAYg9nKBeKBvKB9DpIDsCGAthAFxQBEAEvgM4CC5ANFPmTcAE4CWuA5lAF8oAHzQCAUOIDGcXOygAzBMjEsasBJK4KoACiVwAdFhwQCxZEhRVaDAJRpxUZ4oSH84ieNNQAQvg4VdBNwMyJSCmp6JhY2Th5+IVFg7FDzCPIAUi5eXDgOaGAACy4aTPJBSRk5YCgAIwCgoVo-AK0dXQaOY1S8cMtrKPtHF3qA909xIA

Given that, the next commit implements the Relay strategy again.

AlecAivazis: Ahhh yea okay cool that makes total sense. thanks for putting the playground together 👍

AlecAivazis (approved):

PR #1016, opened 2023-03-22
Blocking strategy

### LINKED ISSUE #1013: Add `@blocking` directive and config
### Describe the feature

Many users have been asking for a way to set `blocking` to `true` globally for their application. We should provide configuration values to set a default behavior as well as directives (ie `@blocking`) for configuring the behavior when relying on a generated load function.

### Criticality

### LINKED ISSUE #1015: Wrong link from docs:
### Describe the bug

Wrong link from docs:

```
The 'HoudiniGraphql/houdini' repository doesn't contain the 'packages/houdini/src/runtime/client/DocumentStore.ts' path in 'main'. 
```

image

### Severity

annoyance

Go here: https://houdinigraphql.com/api/client

image

### Reproduction

_No response_

jycouet:
Fixes #1013, Fixes #1015

### Reminder
- [x] if throwOnError includes all or query, blocking needs to be true
- [x] add in config: default blocking strategy
- [x] add directives `@blocking` `@no_blocking`
  - [x] can't have both
- [x] add a warning if a config is "wrong"
  - [x] having a throwOnError(all|query) & default blocking is set to false
  - [x] Just document that `@blocking_false` will never throwOnError
- [x] update doc query/manual-loading/last chapter about load & result section


AlecAivazis (changes requested):

packages/houdini/src/codegen/generators/artifacts/index.ts
```diff
@@ -365,6 +366,32 @@ export default function artifactGenerator(stats: {
 							artifact.policy = config.defaultCachePolicy
 							artifact.partial = config.defaultPartial
 						}
+
+						// blocking strategy
+						// when can it be true?
+						// 1. default blocking always_blocking & no no_blocking directive
+						// 2. default blocking not_always_blocking & blocking directive
+						if (config.defaultBlockingMode === 'always_blocking') {
```
AlecAivazis, line 374: Reminder that this should go into `houdini-svelte`

packages/houdini/src/runtime/client/documentStore.ts
```diff
@@ -107,6 +107,7 @@ export class DocumentStore<
 		cacheParams,
 		setup = false,
 		silenceEcho = false,
+		blocking,
```
AlecAivazis, line 110: If we check for the conflicting settings inside of the `fetch` method of query store we can avoid adding `blocking` here

AlecAivazis (changes requested):

packages/houdini/src/codegen/generators/artifacts/index.ts
jycouet: We have to keep in mind that, at some point, some frameworks will share some behavior.
For now, I switch it to svelte 💪^^

jycouet:
😅
```mermaid
flowchart TD
    A(Config) --> AB[blocking]
    A --> AN[no blocking]
    AB --> B(throwOnError settings)
    AN --> B
    B --> BB[blocking]
    B --> BN[no blocking]
    BB --> C(blocking directives)
    BN --> C
    C --> CB[blocking]
    C --> CN[no blocking]
    CB --> D(fetch args)
    CN --> D
    D --> DB[blocking]
    D --> DN[no blocking]
    DB --> E(Need to block?)
    DN --> E
```

packages/houdini-svelte/src/plugin/artifactData.ts
```diff
@@ -20,15 +26,44 @@ export function artifactData({
 			if (node.name.value === config.loadDirective) {
 				isManualLoad = false
 			}
+
+			// look for the operation
+			const operations = document.document.definitions.filter(
```
AlecAivazis, line 31: Most of this block isn't necessary. The whole point of `graphql.visit` is to do this search for you. The function passed to `Directive` is only called when the visitor encounters a directive: you just have to check if  `node.name.value === config.blockingDirective`.

packages/houdini-svelte/src/plugin/index.ts
```diff
@@ -84,6 +84,20 @@ export const redirect = svelteKitRedirect
 	// custom logic to pull a graphql document out of a svelte file
 	extractDocuments: extract,
 
+	schema({ config }) {
```
AlecAivazis, line 87: While we're here, can we move the load directives into the plugin (since they're not used by react)

packages/houdini-svelte/src/plugin/index.ts
```diff
@@ -223,6 +237,11 @@ export type HoudiniSvelteConfig = {
 	 */
 	client?: string
 
+	/**
+	 * Specifies whether the default blocking mode is blocking or not. (default: `not_always_blocking`)
+	 */
+	defaultBlockingMode?: 'not_always_blocking' | 'always_blocking'
```
AlecAivazis, line 243: i think `true | false` is a fine value here. Also, can we name it something like `defaultRouteBlocking`?

packages/houdini/src/codegen/generators/artifacts/index.ts
AlecAivazis: I think the react bindings are a great example where the equivalent of `@load` is done with framework-specific tools instead of what we do in svelte. We can always bring them back if/when we need them

packages/houdini-svelte/src/plugin/transforms/kit/load.ts
```diff
@@ -300,7 +300,9 @@ function add_load({
 							AST.objectProperty(AST.literal('event'), AST.identifier('context')),
 							AST.objectProperty(
 								AST.literal('blocking'),
-								AST.booleanLiteral(afterLoad || on_error)
```
AlecAivazis, line 303: This seems strange - was there an issue with passing `false`?

packages/houdini-svelte/src/plugin/validate.test.ts
```diff
@@ -71,6 +71,24 @@ const table: Row[] = [
             `,
 		],
 	},
+	{
+		title: '@blocking @no_blocking on a query',
+		pass: false,
+		documents: [
```
AlecAivazis, line 77: These tests need 2 documents each so that we verify that we group error messages together

packages/houdini/src/runtime/client/plugins/throwOnError.ts
```diff
@@ -25,6 +29,21 @@ export const throwOnError =
 			}[kind]
 
 		return {
+			async start(ctx, { next }) {
+				// add a warning if the config is wrong
+				if (operations.includes('all') || operations.includes('query')) {
+					// if explicitly set to not_always_blocking, we can't throw, so warn the user.
+					const config_not_always_blocking =
+						(ctx.config.plugins as any)['houdini-svelte']?.defaultBlockingMode ===
```
AlecAivazis, line 37: `houdini` shouldn't know anything about `houdini-svelte`. Can we move this into the `fetch` method with the actual `blocking` logic?

packages/houdini/src/runtime/client/index.ts
```diff
@@ -44,6 +43,9 @@ export class HoudiniClient {
 	// the list of plugins for the client
 	#plugins: ClientPlugin[]
 
+	// expose operations settings
+	throwOnError_operations: ThrowOnErrorOperations[]
```
AlecAivazis, line 47: Can we mark this as `readonly`?

site/src/routes/api/query/+page.svx
```diff
@@ -295,23 +295,12 @@ export const load: PageLoad = async (event) => {
...
-	const MyQuery = new MyQueryStore()
-
-	const result = await MyQuery.fetch({ event })
-
-	console.log('do something with', result)
-
-	return { MyQuery }
-}
-```
+- Use the `blocking: true` parameter in the `fetch` method _(you can check the deep dive "Blocking Navigation" bellow)_
```
AlecAivazis, line 301: We need to add a subsection in the automatic loading section that talks about blocking the generated queries. Maybe we frame it around "Loading States"?

site/src/routes/api/graphql-magic/+page.svx
```diff
@@ -81,6 +81,14 @@ mutation NewItem($input: AddItemInput!) {
 
 A field marked with `@paginate` is updated with calls to the page loaders returned by the pagination function. For more information on pagination check out the [guide](/guides/pagination). If you pass a value for the `name` argument, the underlying list can be updated using the [operation fragments](/api/mutation#lists).
 
+### `@blocking`
+
+A query marked with `@blocking` will always await the fetch.
+
+### `@no_blocking`
```
AlecAivazis, line 88: I still prefer `@unblocking` (or just `@unblock`/`@block` ). the `_` casing has been reserved for very special cases in the public API so I think we should try to avoid it unless we have a _very_ good reason

AlecAivazis (changes requested):

packages/houdini-svelte/src/plugin/index.ts
jycouet: With `BlockingMode` it was too open; that's why I had `'not_always_blocking' | 'always_blocking'` _(that I didn't like)_
Narrowing it down makes more sense 👍
`defaultRouteBlocking`? or should we express that it's client-side navigation only?

packages/houdini-svelte/src/plugin/transforms/kit/load.ts
jycouet: Yes, setting `false` explicitly says at args level: "I don't want it", but it shouldn't be the case.
Here, `afterLoad || on_error` we want to explicitly say YES otherwise we don't express anything. Like this `@directives` can overwrite the situation.

=> Locally in a app we can have a different behavior
=> In e2e tests we can do all situations like this

site/src/routes/api/query/+page.svx
jycouet: And removing the `deep dive: Blocking navigation`? Or in addition?

In the end, the blocking concept (whatever the name) applies in Automatic & Manual.
Should we add a chapter after `Manual Loading` with the name `Loading States`?

site/src/routes/api/graphql-magic/+page.svx
jycouet: Let's go with `@unblocking` 👍

packages/houdini-svelte/src/plugin/index.ts
AlecAivazis: I dont think most people really _care_ that it happens one one versus the other. Also, its very possible that one day we start to do something clever on the backend and resolve early if its taking too long

packages/houdini-svelte/src/plugin/transforms/kit/load.ts
AlecAivazis: Ahh okay cool 👍

packages/houdini/src/codegen/validators/typeCheck.ts
```diff
@@ -1103,6 +1088,40 @@ function checkMaskDirective(config: Config) {
 	}
 }
 
+function checkBlockingDirectives(config: Config) {
```
AlecAivazis, line 1091: This should also move into `houdini-svelte` (along with `config.blockDirective`).

packages/houdini/src/runtime/client/plugins/fetch.ts
```diff
@@ -32,31 +32,61 @@ export const fetch = (target?: RequestHandler | string): ClientPlugin => {
...
+					resolve(ctx, {
+						fetching: false,
+						variables: ctx.variables ?? {},
+						data: result.data,
+						errors: !result.errors || result.errors.length === 0 ? null : result.errors,
+						partial: false,
+						stale: false,
+						source: DataSource.Network,
+					})
+				} catch (error: unknown) {
```
AlecAivazis, line 61: if the error isn't the correct type, this will never resolve with the error which I dont think is the desired behavior. A user would encounter this error when they fetch something which means that we should treat these errors like normal network requests errors (which can be arrays)

site/src/routes/api/config/+page.svx
```diff
@@ -69,10 +69,11 @@ export default {
 
 Here is a summary of the possible configuration values:
 
-- `client` (optional, default: `./src/client`): a relative path (from houdini.config.js) to a file that exports your houdini client as its default.
+- `client` (optional, default: `"./src/client"`): a relative path (from houdini.config.js) to a file that exports your houdini client as its default.
+- `defaultRouteBlocking` (optional, default: `false`): Specifies whether the client side routing is blocking or not.
```
AlecAivazis, line 73:
```suggestion
- `defaultRouteBlocking` (optional, default: `false`): Specifies the default blocking behavior for client-side navigation. For more information, please visit [this section of the docs](https://houdinigraphql.com/api/query#manual-loading).
```

packages/houdini-svelte/src/plugin/artifactData.ts
```diff
@@ -1,7 +1,11 @@
 import * as graphql from 'graphql'
 import type { Document, Config } from 'houdini'
 
-export type PluginArtifactData = { isManualLoad?: boolean }
+export type PluginArtifactData = {
+	isManualLoad?: boolean
+	set_blocking?: boolean
```
AlecAivazis, line 6: Is there a reason to have 2 different keys in the object? They are mutually exclusive, right? Would a single `boolean | undefined` not work?

packages/houdini/src/runtime/client/plugins/fetch.ts
jycouet: I removed this.
We will think about this in another PR

packages/houdini-svelte/src/plugin/artifactData.ts
jycouet: 🤔 We can have a load that is blocking or not blocking.
I would say that it's not exactly the same thing.

packages/houdini/src/codegen/index.ts
```diff
@@ -117,7 +117,8 @@ export async function runPipeline(config: Config, docs: Document[]) {
 			docs
 		)
 	} catch (e) {
-		error = e as Error
+		// flatten the error array
+		error = Array.isArray(e) && e.length === 1 ? e[0] : e
```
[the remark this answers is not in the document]
jycouet: Not sure how to address this.
Why doing something only for testing environment?

Let's speak later

AlecAivazis (approved):
Nice job!

PR #1019, opened 2023-03-28
Bug/include directive null

### LINKED ISSUE #1018: include or skip directive that excludes a field cause the data to be set to null
### Describe the bug

Adding `@include(if: false)` or `@skip(if: true)` on a query, causes the entire `store.data` to be set to null, instead of only that single field. This was working fine in `1.0.11`, but seems to be broken when we upgraded to `1.1.3`.

example query:
```gql
query Test {
  user(id: "1") {
    name
  }
  hello @skip(if: true)
}
```

Now, `$store.data` will be `null`, instead of the expected `{"user":{"name":"some name"}}`

I haven't tested this with adding the include/skip directive on a field inside of the `user` object, but I think that is still working fine.

### Severity

blocking an upgrade

1. Check out the commit in the linked reproduction
2. Run the e2e kit sample and go to /stores/directives
3. See the data on the page being `null`
4. Set `@skip(if: false)` and `@include(if: true)`
5. See the page working now

### Reproduction

https://github.com/SeppahBaws/houdini/commit/70f1af8353fb174eba9f4db90cb7c62d93ce1159

AlecAivazis:
Fixes #1018 

This PR fixes a bug in the cache behavior that was ignoring `include` and `skip` directives causing a `null` value to incorrectly cascade.

### DIFF
--- a/packages/houdini/src/runtime/cache/cache.ts
+++ b/packages/houdini/src/runtime/cache/cache.ts
@@ -828,11 +828,32 @@ class CacheInternal {
 		// look at every field in the parentFields
 		for (const [
 			attributeName,
-			{ type, keyRaw, selection: fieldSelection, nullable, list, visible },
+			{ type, keyRaw, selection: fieldSelection, nullable, list, visible, directives },
 		] of Object.entries(targetSelection)) {
 			// skip masked fields when reading values
 			if (!visible && !ignoreMasking) {
 				continue
+			}
+
+			// some directives like @if and @include prevent the value from being in the
+			// selection
+			const includeDirective = directives?.find((d) => {
+				return d.name === 'include'
+			})
+			if (includeDirective) {
+				// if the `if` argument evaluates to false, skip the field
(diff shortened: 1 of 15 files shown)


packages/houdini/src/runtime/cache/cache.ts
```diff
@@ -828,13 +828,34 @@ class CacheInternal {
...
 			attributeName,
-			{ type, keyRaw, selection: fieldSelection, nullable, list, visible },
+			{ type, keyRaw, selection: fieldSelection, nullable, list, visible, directives },
 		] of Object.entries(targetSelection)) {
 			// skip masked fields when reading values
 			if (!visible && !ignoreMasking) {
 				continue
 			}
 
+			// some directives like @if and @include prevent the value from being in the
```
SeppahBaws, line 838: Small mistake in the comment 🙂 
```suggestion
			// some directives like @skip and @include prevent the value from being in the
```

AlecAivazis: Thanks!

PR #1027, opened 2023-03-30
Fix bug when fragment arguments are nested in objects

AlecAivazis:
This MR fixes an issue reported in discord by @SeppahBaws that was causing issues when a fragment argument was nested inside of an object


packages/houdini/src/codegen/generators/artifacts/index.ts
```diff
@@ -232,6 +227,9 @@ export default function artifactGenerator(stats: {
 						const matchingFragment = fragments.find(
 							(fragment) => fragment.name.value === name
 						)
+						if (name.includes('UserSearch')) {
+							console.log('searching `name.value`', name, ':', originalString)
+						}
```
SeppahBaws, line 232: Ah this was some code from my testing, can be removed

AlecAivazis: Nice catch!

PR #1033, opened 2023-03-31
Handle `null` in `watchSchema.interval` properly

devunt:
Since JS [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) treats `null` as `nullish` value, passing over `null` in `watchSchema.interval` forcefully makes the value to `2000`, not `null`.

Houdini handles `2000`, `0`, and `null` differently in its `watchSchema.interval` config, but this bug prevents users from using `null` as its value to dictate the behavior to never pull the schema on any circumstances.

This patch fixes that behavior.

### DIFF
--- a/.changeset/tender-ducks-cough.md
+++ b/.changeset/tender-ducks-cough.md
@@ -1,0 +1,5 @@
+---
+'houdini': patch
+---
+
+Fix passing `null` over `watchSchema.interval` in the configuration does not work as expected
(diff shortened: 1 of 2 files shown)


AlecAivazis (approved):
Nice catch! Thanks for the PR 👍

PR #1048, opened 2023-04-10
Include explicit `types` export conditions in package.json `exports`

### LINKED ISSUE #1046: Errors when using `nodenext`, `node16`, or `bundler` moduleResolution in tsconfig.json
### Describe the bug

when using the newer TypeScript moduleResolution options `nodenext`, `node16`, or `bundler` (avail in TS 5), can resolve type declaration files. Getting the following error in `vite.config.ts` when I change the tsconfig.json moduleResolution to one of the previous mentioned options.

` Error: Could not find a declaration file for module 'houdini/vite'.`

This is because the `exports` in package.json does not explicitly definite the `types` conditional export and falls back to the default location which doesn't exist.

This is how the `exports` section is define currently in package.json

```
"exports": {
    "./package.json": "./package.json",
    "./codegen": {
      "import": "./build/codegen-esm/index.js",
      "require": "./build/codegen-cjs/index.js"
    },
    ".": {
      "import": "./build/lib-esm/index.js",
      "require": "./build/lib-cjs/index.js"
    },
    "./test": {
      "import": "./build/test-esm/index.js",
      "require": "./build/test-cjs/index.js"
    },
    "./vite": {
      "import": "./build/vite-esm/index.js",
      "require": "./build/vite-cjs/index.js"
    }
  },
```

But because there is no explicit `types` export, ts will look for the type definition name `./build/vite-esm/index.d.ts` which doesn't exist.  it looks like you were trying to do this with the `typesVersion` section but TS only looks at `exports` when the moduleResolution is `nodenext`, `node16`, or `bundler`. 

The correct way would be to define it as follows:

```
"exports": {
    "./package.json": "./package.json",
    "./codegen": {
      "types": "./build/codegen/index.d.ts",
      "import": "./build/codegen-esm/index.js",
      "require": "./build/codegen-cjs/index.js"
    },
    ".": {
      "types": "./build/lib/index.d.ts"
      "import": "./build/lib-esm/index.js",
      "require": "./build/lib-cjs/index.js"
    },
    "./test": {
      "types": "./build/test/index.d.ts",
      "import": "./build/test-esm/index.js",
      "require": "./build/test-cjs/index.js"
    },
    "./vite": {
      "types": "./build/vite/index.d.ts",
      "import": "./build/vite-esm/index.js",
      "require": "./build/vite-cjs/index.js"
    }
  },
```

I have patched via pnpm patch using the above changes and all is good with the project again. But would be ideal to get this into the codebase.  Happy to push up a PR if that would be preferred.

### Severity

annoyance

1. Edit tsconfig.json - add `"moduleResoultion": "bundler"` under `compilerOptions`
2. Open up vite.config.ts - should notice that the `import houdini from 'houdini/vite'` is show in error

```
TS7016: Could not find a declaration file for module 'houdini/vite'. '.../node_modules/.pnpm/houdini@1.1.7/node_modules/houdini/build/vite-esm/index.js' implicitly has an 'any' type.   There are types at '.../node_modules/houdini/build/vite/index.d.ts', but this result could not be resolved when respecting package.json "exports". The 'houdini' library may need to update its package.json or typings.
```

### Reproduction

_No response_

mpellegrini:
Fixes #1046

Include explicit `types` export conditions in
package.json `exports`.  The d.ts files are located elsewhere so can't relay on default value which is the same as the import, except with a `d.ts` extension.

### DIFF
--- a/.changeset/rude-socks-doubt.md
+++ b/.changeset/rude-socks-doubt.md
@@ -2,7 +2,6 @@
 'houdini-plugin-svelte-global-stores': patch
 'houdini-svelte': patch
 'houdini-react': patch
-'scripts': patch
 'houdini': patch
 ---
 
--- a/.changeset/rude-socks-doubt.md
+++ b/.changeset/rude-socks-doubt.md
@@ -1,0 +1,9 @@
+---
+'houdini-plugin-svelte-global-stores': patch
+'houdini-svelte': patch
+'houdini-react': patch
+'scripts': patch
+'houdini': patch
+---
+
+Include explicit types export conditions in package.json exports
--- a/packages/_scripts/build.js
+++ b/packages/_scripts/build.js
@@ -1,9 +1,9 @@
 import esbuild from 'esbuild'
 import { replace } from 'esbuild-plugin-replace'
-import fs from 'fs/promises'
 import glob from 'glob'
-import path from 'path'
-import { promisify } from 'util'
+import fs from 'node:fs/promises'
+import path from 'node:path'
+import { promisify } from 'node:util'
 
 // the relevant directories
 const build_dir = path.join(process.cwd(), 'build')
@@ -38,6 +38,7 @@ export default async function ({ plugin }) {
 			// when there's a plugin directory, that is the main entry point
 			package_json.main = './build/plugin-cjs/index.js'
 			package_json.exports['.'] = {
+				types: './build/plugin/index.d.ts',
 				import: './build/plugin-esm/index.js',
 				require: './build/plugin-cjs/index.js',
 			}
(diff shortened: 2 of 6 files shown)


AlecAivazis:
Awesome! Thanks for putting this together.

Mind adding a changeset? If you include this as a `patch` I can release it before the upcoming 1.2 release

mpellegrini:
I have added a patch changeset. Hopefully I did this as expected. If not let me know and will update accordingly.  Thanks for your help @AlecAivazis

.changeset/rude-socks-doubt.md
```diff
@@ -0,0 +1,9 @@
+---
+'houdini-plugin-svelte-global-stores': patch
+'houdini-svelte': patch
+'houdini-react': patch
+'scripts': patch
```
AlecAivazis, line 5: Since the `scripts` package is marked `private` it shouldn't be included in here. 

```suggestion
```

AlecAivazis (changes requested):
Overall looks pretty good! One small tweak and we can merge this in

AlecAivazis (approved):

PR #1049, opened 2023-04-11
Send `operationName` along with the `query` and `variables` by default

devunt:
Rationale: #832

This PR changes the default behavior of `fetch` built-in plugin to include `operationName` in the fetch payload along with the `query` and `variables`.

This PR also changes the signature of `subscription` built-in plugin to pass `operationName` to user-provided subscription function so the user can easily interacts with it in their implementation.
Several popular implementations such as `graphql-ws` and `graphql-sse` do already utilize `operationName` argument if it is passed into their `subscribe` function. So in the most cases if we pass `operationName` to `subscribe` function it would be automatically included in the corresponding payload, which is also good for interoperability.

This also makes Houdini aligned with other popular graphql client implementations such as Apollo, Relay, Urql. All of them are sending `operationName` to the server by default, and it can not be disabled by user as it is a de facto standard to include it and have no benefit to disable it except for saving few network bytes. (although it is a not a hard requirement [according to the spec](https://graphql.org/learn/serving-over-http/#post-request), if the `query` contains only single operation.)

### DIFF
--- a/.changeset/twenty-months-cover.md
+++ b/.changeset/twenty-months-cover.md
@@ -1,0 +1,5 @@
+---
+'houdini': patch
+---
+
+Send `operationName` along with the `query` and `variables` by default.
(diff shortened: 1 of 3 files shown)


AlecAivazis (approved):
Makes total sense! Thanks for the thoughtful contribution 🙌

PR #1050, opened 2023-04-11
Centralize JS code parsing/printing mechanism into core

### LINKED ISSUE #1047: Houdini reports an error when import asseertion is used in the code
### Describe the bug

Houdini reports an error when import assertion is used in the code, even in the totally not related code.
I suspect there's an annoying bug with AST parser.

### Severity

annoyance

1. Create `src/lib/test.json` file with `{}` as its only contents (contents here really doesn't matter)
2. Create `src/lib/test.ts` file with following code and save it.
    ```typescript
    import obj from './test.json' assert { type: 'json' };
    ```
3. Vite dev server throws `❌ Encountered error in src/lib/test.ts`, even if `test.ts` is used in nowhere and `obj` is also not used.

### Reproduction

_No response_

devunt:
Fixes #1047

Recast also provides set of parsing JS/TS code functions that built on top of `@babel/parser`. This PR changes JS/TS code parsing mechanism from using `@babel/parser` to using recast's one, so we can get benefit from its [predefined babel options](https://github.com/benjamn/recast/blob/master/parsers/_babel_options.ts).

Since recast expects the same instance is used between parsing and printing, it is required to use the same recast instance among the whole codebase. This PR makes `houdini` package to export the `printJS` function in along with existing `parseJS` function and make plugins to import it when code printing is needed.

Some test snapshots were also updated, as recast now uses its original context to render the resulting code. This wasn't available before as the parsing was done via `@babel/parser`, not recast. The only changes to snapshots are whitespaces and code styles.


packages/houdini/src/lib/index.ts
```diff
@@ -18,3 +18,5 @@ export type { EmbeddedGraphqlDocument } from './walk'
 
 export * as fs from './fs'
 export * as path from './path'
+
+export * as recast from 'recast'
```
AlecAivazis, line 22: I'm not a huge fan of exporting `recast` from the core houdini module just to save the one import line. Especially since it's very likely I will forget 😅 

I think if we were thinking about abstracting the `recast` dependency away from the rest of the codebase I think it would make sense exported as a different name (something I wouldn't be opposed to exploring in a separate PR) but as it is now, I don't think there's a lot of value added to justify the cost of confusing people who aren't confident about the difference.

AlecAivazis (changes requested):

packages/houdini/src/lib/index.ts
devunt: As like stated in the PR description, this has to be done because `recast` expects that the same instance is used between its `parse` and `print` function. Plugins were using `parseJS` from the houdini core, and used their own instance of `recast` to `print` it to the actual code, which would raise an error in `print`.

Btw, I also agree with re-exporting whole `recast` from the houdini core is not a great idea. I think exporting something like `printJS` would be more appropriate, so making sure the same instance is used between `parse` and `print`, by using corresponding `parseJS` and `printJS` functions from the houdini core.

AlecAivazis: ohhh so sorry I totally missed that in your description. Are we still able to use the AST builders from a different instance? If so, I think adding `printJS` to `lib` is a great solution. If we have to use the builders from the same instance than we probably want to come up with a good name for those too

devunt: Type building is done via `ast-types` package, and they don't have any assertion on instance equality checking (I looked over the code).

Detailed explanation:

The exact problematic line is [here](https://github.com/benjamn/recast/blob/5e6442d2d66bb00af581b8d08b2ac970bd6d4a5c/lib/patcher.ts#L28), which ensures that the `lines` argument are the instance of `Lines` [class](https://github.com/benjamn/recast/blob/5e6442d2d66bb00af581b8d08b2ac970bd6d4a5c/lib/lines.ts#L25) by using `lines instanceof Lines`. They grab `lines` [from passed code's](https://github.com/benjamn/recast/blob/5e6442d2d66bb00af581b8d08b2ac970bd6d4a5c/lib/patcher.ts#L172)  `loc.lines` field which [was generated by](https://github.com/benjamn/recast/blob/5e6442d2d66bb00af581b8d08b2ac970bd6d4a5c/lib/parser.ts#L15) `parse` function (`loc` was `undefined` when using `@babel/parser` since they don't populate that field, thus making `recast.print` not trying to validate `loc.lines` at all, but now we have them).

if we call `recast.parse` and `recast.print` from different packages, even if they are on the same version, bundler treats it differently and creates two identical `Lines` class which will be like `class Lines_1` and `class Lines_2`, effectively making `parse` returns the object with `Line_1` and `print` try to validate that the object is instance of `Line_2`.

.changeset/green-camels-walk.md
```diff
@@ -0,0 +1,7 @@
+---
+'houdini-svelte': minor
```
AlecAivazis, line 2: Since this PR doesn't add a new feature to Houdini it should be classified as a `patch`

packages/houdini-svelte/src/plugin/transforms/componentQuery.test.ts
```diff
@@ -23,9 +23,7 @@ test('no variables', async function () {
 		import { getCurrentConfig } from "$houdini/runtime/lib/config";
 		import { marshalInputs } from "$houdini/runtime/lib/scalars";
 		const _houdini_TestQuery = new TestQueryStore();
-
-		$:
-		value = _houdini_TestQuery;
+		$: value = _houdini_TestQuery
```
AlecAivazis, line 26: wow it is silly how happy this change makes me. it always bothered me for absolutely no good reason at all

AlecAivazis (changes requested):

packages/houdini-svelte/src/plugin/transforms/kit/init.test.ts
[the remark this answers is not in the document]
devunt: My bad. It was serializing the whole parsing result instead of just snapshot-ing the resulting human-readable code. Fixed and snapshots are updated.

AlecAivazis (approved):

devunt:
Rebased from `main` due to the conflict introduced by #1016

AlecAivazis (approved):
Thanks for rebasing!

PR #1053, opened 2023-04-13
fix: Duplicate virtual `+layout.js` files

knd775:
Duplicate `+layout.js` files were being created in the root directory. This was an issue without side effects until sveltekit 1.15.5, which added a duplicate check. This fix allows houdini to work in sveltekit again.

### DIFF
--- a/.changeset/mighty-rice-trade.md
+++ b/.changeset/mighty-rice-trade.md
@@ -1,0 +1,5 @@
+---
+'houdini-svelte': patch
+---
+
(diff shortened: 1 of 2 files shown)


AlecAivazis (changes requested):
Thanks for getting this together! Mind running prettier really quickly?

knd775:
> Thanks for getting this together! Mind running prettier really quickly?

Done!

AlecAivazis (approved):

PR #1081, opened 2023-05-05
Add Router to React Bindings

AlecAivazis:
This PR adds a very experimental router to the react integration and sets the ground work for a react-based framework for building GraphQL applications 👀


packages/houdini/src/runtime/client/plugins/cache.ts
```diff
@@ -18,6 +18,11 @@ export const cachePolicy =
 		serverSideFallback?: boolean
 	}): ClientPlugin =>
 	() => {
+		// if we were given a cha
+		if (cache) {
```
AlecAivazis, line 22: 🙅

Issue #1066, opened 2023-04-21
Queries affect stores of other queries


knd775:
### Describe the bug

We have three queries that seem to be interacting with each other somehow. We have a list of objects that we display in a table. It loads correctly. When we click on a row, it opens a modal that runs another query to get more data about the object in that row. As soon as this query returns, some of the overlapping data disappears from the table. Then, when we hover over the table (which triggers a tooltip that loads some of the now missing data), the data pops back in the table. These queries (and their stores) should be entirely separate from each other, but that doesn't seem to be the case here.

This only happens when there is some overlap between the objects being requested.

Discussed here:
https://discord.com/channels/1024421016405016718/1099015262310777005

### Severity

blocking all usage of Houdini

### Steps to Reproduce the Bug

Using repro below
- Run `pnpm i`, then `pnpm dev`
- Click one of the run buttons on the page, and see that data is being updated for the query that is not run

### Reproduction

https://github.com/knd775/houdini-bug-repros/tree/main

PR #1085, opened 2023-05-08
Fix Loading - without load in an external file

jycouet:
Fixes [Discord](https://discord.com/channels/1024421016405016718/1104034400020213821/1104034400020213821)


SeppahBaws:
Heya 👋
Any ETA on this? We'd like to push another release to prod but we can't roll back to 1.2.0 since we rely on the bugfix in 1.2.2.

AlecAivazis:
Finding the right heuristic to always have the correct initial value is proving to be kind of tricky. In order to unblock your deploy @SeppahBaws I propose that we add a `fetching` argument to the query store constructor that you can use to set the initial value. I know it's not ideal but it seems like something that we would always regardless of having a smart default and it removes some of the pressure so we can think carefully here. How does that sound @jycouet?

jycouet:
I'm not really in favor of a `fetching` argument in the query store constructor.
- It's introducing a new thing for users and putting the responsibility on them.
- It will be hard to remove. (Breaking change?)

---

This PR is trying to fix a regression from https://github.com/HoudiniGraphql/houdini/pull/1068 _(that fixed https://github.com/HoudiniGraphql/houdini/issues/1066)_

Right now, the PR
- Fixes The `fetching` issue
- Keep https://github.com/HoudiniGraphql/houdini/issues/1066 working
- But, a new problem showed up: https://github.com/HoudiniGraphql/houdini/blob/667b8407e1090ba004787506e7d5410a31e5b1b2/e2e/kit/src/routes/stores/mutation-update/+page.svelte#L8-L10 I will dig on it this morning.

---

@SeppahBaws, to unblock you, can't you put the query in the Svelte file directly? So that you can roll out your release?
Let us know.

SeppahBaws:
> @SeppahBaws, to unblock you, can't you put the query in the Svelte file directly? So that you can roll out your release?
That would work as a workaround, but in some of our components we mix and match up to 5 different query stores, which would result in adding +-200 extra LOC in that component, something I'm a bit reluctant to do.

I'll have a stab at rewriting the logic in these components to compensate for the fact that those stores will be `fetching = true` by default, I'll let you know how it goes.

SeppahBaws:
Update:
The fix @jycouet shared with me in Discord did the trick, without bloating my component file 🥳 

### For anyone else running into the same issue:

If you have query file like this: `MyQuery.gql`:
```gql
query MyQuery {
  me {
    name
  }
}
```
simply change it to this: `MyQuery.svelte`:
```svelte
<script lang="ts">
  import { graphql } from "$houdini";
  graphql(`
    query MyQuery {
      me {
        name
      }
    }
  `);
</script>
```

Since it's now in a svelte file, the store will have the correct `fetching` state by default, and I don't have to mess around in my component itself :)

PR #1090, opened 2023-05-18
Persisted Queries

jycouet:
Fixes
- Artifact Hash and Hashes in the generated file are now matching

Adding
- Persisted Queries File can be generated on each changes via `PersistedQueriesPath` param in the config file.


.changeset/fresh-spiders-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+'houdini': patch
+---
+
+fix: Artifact Hash and Hashes in the generated file are now matching
```
AlecAivazis, line 5: I don't think this is the case anymore, right?

.changeset/happy-peas-tickle.md
```diff
@@ -0,0 +1,5 @@
+---
+'houdini': patch
+---
+
+feat: Persisted Queries File can be generated on each changes via `PersistedQueriesPath` param in the config file.
```
AlecAivazis, line 5: Not totally related to this exact change but i wanted a place to start a thread: What if we just always generated the mapping? All the config option would do is configure where it ends up

jycouet: Yeah, while working on it, I was thinking about that too. I don't really have an opinion on it.
Do you think it makes more sense?

.changeset/fresh-spiders-behave.md
jycouet: If there is one thing in the PR it's this. _Let me try to explain._

Let's say, I have a query: 
```gql
query TestQuery1 { version }
```
Houdini  generates an artifact: 
```js
{
  name: 'TestQuery1',
  hash: '49dc4c0692eced5aacee9f00666346a949191deb9d6efbaed8f073f1dc7e16fd',
  // ...
}
```

Then today, you generate the file with: `npx houdini generate --output ./queries.json`
Looking at the generated file `./queries.json` you will see:
```js
{
  '361432f464ed44eed788f3ea66c4dabc46437b88edbe7daccca87045fd31447f': 'query TestQuery1 {\n  version\n}\n'
}
```
As the 2 hashes don't match, when the client send `49dc4c0692eced5aacee9f00666346a949191deb9d6efbaed8f073f1dc7e16fd`, the server doesn't find the query.

I added these tests https://github.com/HoudiniGraphql/houdini/pull/1090/files#diff-75a9bc287c2c9ef7991af013fae5e2db6560e74ee39843d1fcb5b2e03873fcaeR51 to make sure that `artifact.hash` finds the query in the generated json file.

Does it make sense? Or I missed something?

.changeset/happy-peas-tickle.md
AlecAivazis: Sorry for not responding earlier! Yea, I think it simplifies the overall story and implementation a lot

jycouet: After prettier, lint & other... It's done ^^
haha

AlecAivazis (approved):

AlecAivazis (approved):

PR #1093, opened 2023-05-19
Node interface warning update

jycouet:
- Node interface message update (trying to give more info)
- Message is now not displayed if the Node interface is not used
- Message will show on usage & wrong configuration

Example of message:
image

### DIFF
--- a/packages/houdini/src/lib/config.ts
+++ b/packages/houdini/src/lib/config.ts
@@ -389,10 +389,17 @@ export class Config {
 	}
 
 	keyFieldsForType(type: string) {
+		// TODO JYC: any special reason for this? (in postegraphile, there is a node interface on query)
+		// """
+		// Exposes the root query type nested one level down. This is helpful for Relay
+		// which can only query top level fields if they are in a particular form.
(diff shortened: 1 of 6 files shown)


packages/houdini/src/lib/config.ts
```diff
@@ -389,10 +389,17 @@ export class Config {
...
+		// """
+		// query: Query!
+
 		// the only type that doesn't have a well defined ID is the root query type
-		return this.schema.getQueryType()?.name === type
-			? []
-			: keyFieldsForType(this.configFile, type)
+		return keyFieldsForType(this.configFile, type)
+		// return this.schema.getQueryType()?.name === type
+		// 	? []
```
jycouet, line 402: it seems to be on purpose that `Query` was left out (with the comment). @AlecAivazis do you remember why?

I added a comment on top. Let's chat 👍
(Not urgent)

AlecAivazis: The key for the `Query` type is the static `__ROOT__` value so there are no keys

PR #1098, opened 2023-05-26
filepath ignoring QueryString

jycouet:
Fixes [#discord](https://discord.com/channels/1024421016405016718/1111684053880553482/1111780792033628170)

---

FYI @Lms24
- This PR is to remove `?sentry-auto-wrap`
- Also, just to confirm, `houdini` needs to come before `sentrySvelteKit` in the vite config (to transform file)

### DIFF
--- a/packages/houdini/src/lib/config.ts
+++ b/packages/houdini/src/lib/config.ts
@@ -473,11 +473,6 @@ export class Config {
 		return false
 	}
 
-	private pathRmvQueryString(filepath: string) {
-		const parsed = path.parse(filepath)
-		return `${parsed.dir}/${parsed.name}${parsed.ext.split('?')[0]}`
-	}
-
 	includeFile(
 		filepath: string,
 		{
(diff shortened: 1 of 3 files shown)


packages/houdini/src/lib/config.ts
```diff
@@ -473,13 +473,20 @@ export class Config {
...
+
 	includeFile(
 		filepath: string,
 		{
 			root = this.projectRoot,
 			ignore_plugins = false,
 		}: { root?: string; ignore_plugins?: boolean } = {}
 	) {
+		filepath = this.pathRmvQueryString(filepath)
+
```
AlecAivazis, line 489: Seems like the extra function doesn't really offer us much. How about we just inline the logic like this?

```suggestion
	includeFile(
		filepath: string,
		{
			root = this.projectRoot,
			ignore_plugins = false,
		}: { root?: string; ignore_plugins?: boolean } = {}
	) {
		const parsed = path.parse(filepath)
		filepath =  `${parsed.dir}/${parsed.name}${parsed.ext.split('?')[0]}`

```

jycouet: It was to convey the why, but it's ok. Let's switch

AlecAivazis (approved):

PR #1099, opened 2023-05-27
Refresh query result API docs

rmarscher:
I couldn't remember if `fetching` or `loading` was the current way to check if a query is loading and I couldn't remember if that var exists on the store or on the store value. I went to the [API query docs](https://houdinigraphql.com/api/query)... and they confused me more :)  So I went to the type definition and source code and I think I was able to get it straightened out. Suggesting some updates to that page of the docs:

Here's the current docs on the query store fields:
https://github.com/HoudiniGraphql/houdini/blob/8bba6c3/site/src/routes/api/query/%2Bpage.svx#L91-L94

Here's the current type definition. I think `loading` was replaced by `fetching` a while ago. `variables` seems useful.
https://github.com/HoudiniGraphql/houdini/blob/8bba6c3/packages/houdini/src/runtime/lib/types.ts#L235-L243

I didn't add `stale` and `source` since they seem to be more internal.

I didn't see any docs about how to handle the user wanting to refresh your query. I'm assuming this would be the recommended way?
```
MyQuery.fetch({ variables: $MyQuery.variables, policy: CachePolicy.NetworkOnly });
```

I added a note about that to the query store fetch method.

I also noticed what seems to be a typo here:
https://github.com/HoudiniGraphql/houdini/blob/8bba6c3/site/src/routes/api/query/%2Bpage.svx#LL397C23-L397C23
I think that should be `{#if $MyQuery.fetching}`

### DIFF
--- a/site/src/routes/api/query/+page.svx
+++ b/site/src/routes/api/query/+page.svx
@@ -80,7 +80,7 @@ query MyProfileInfo {
 {$MyProfileInfo.data.viewer.firstName}
 ```
 
-This example takes advantage of Houdini's powerful [load generator](#generating-loaders) and
+This example takes advantage of Houdini's powerful [load generator](#automatic-loading) and
 is just one of many ways to get access to a store that can drive your server-side rendered routes.
 For more information, check out the [Working with GraphQL](/guides/working-with-graphql) guide.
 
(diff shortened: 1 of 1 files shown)


rmarscher:
That example I put in the PR description gives a type error.

``` 
Type 'MyQuery$input | null' is not assignable to type 'MyQuery$input | undefined'.
Type 'null' is not assignable to type 'MyQuery$input | undefined'.
 ```
 
 Need to do this to avoid a type error.
 ```
 MyQuery.fetch({
	variables: $MyQuery.variables || undefined,
	policy: CachePolicy.NetworkOnly
});
```

Maybe that variables input could be update to accept null which would avoid needing to do that? `MyQuery$input | null | undefined`. I could open another issue for that... super minor.

jycouet:
Actually, when you want to refresh values, you can omit `variables`, it will take last ones.
So this should be sufficient:
```ts
MyQuery.fetch({ policy: CachePolicy.NetworkOnly });
```

Could you confirm?

AlecAivazis (dismissed):
Ah whoops! Nice catch 👍

rmarscher:
> Actually, when you want to refresh values, you can omit `variables`, it will take last ones.

Thanks for catching that. I just tested with 1.2.4 and I was able to omit the variables to refresh. That was my expectation for how it should work... but I thought I had experienced needing to explicitly pass the variables to fetch every time in that past (some time in mid to late 2022).

I reverted any update to the fetch docs. :+1:

AlecAivazis (approved):

AlecAivazis:
Thanks @rmarscher!

PR #1103, opened 2023-06-06
Fix inserts into lists of unions

canastro:
Fixes #TICKET


e2e/kit/src/routes/union-insert/InsertA.gql
```diff
@@ -0,0 +1,5 @@
+mutation CreateA($a: String!) {
+  createA(a: $a) {
+    ... All_AorB_insert
```
canastro, line 3: Here is where the compilation issue occurs

jycouet: This should be fixed (I'll tag where)

e2e/kit/houdini.config.js
```diff
@@ -26,6 +26,9 @@ const config = {
   types: {
     RentedBook: {
       keys: ['userId', 'bookId']
+    },
+    UnionAorB: {
+      keys: []
```
jycouet, line 31: Note that Houdini add `keys: ['id']` everywhere. But here, we need to remove it explicitly. 

@AlecAivazis do you think that we should remove the default keys for unions? (I'm not sure what to think here)

e2e/kit/src/routes/union-insert/+page.svelte
```diff
@@ -0,0 +1,33 @@
...
+
+  export let data: PageData;
+
+  $: ({ AorB } = data);
+
+  const addA = graphql(`
+    mutation CreateA($a: String!) {
+      createA(a: $a) {
+        __typename
+        ...All_AorB_insert
```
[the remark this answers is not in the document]
canastro: That seems to work 🎉

e2e/kit/houdini.config.js
canastro: Ah! Thats a tricky for a user of houdini to figure out on their own.

AlecAivazis: Hm, no I think even for unions `id` makes sense as a default. The members of a union should have the field. I would have thought the lack of `id` should have been caught in the type check validator

packages/houdini/src/codegen/generators/artifacts/selection.ts
```diff
@@ -189,6 +189,10 @@ function prepareSelection({
 			let nullable = false
 			if (field.name.value === '__typename') {
 				fieldType = config.schema.getType('String')!
+			}
+			// if it's an union, we are already looking at the correct type
+			else if (graphql.isUnionType(type)) {
+				fieldType = type
```
AlecAivazis, line 195: hm something feels weird about this. 

`type` is the type of the field's parent right? I don't understand why that would always match the parent itself. For example, with a schema like

```graphql
type A {
    field: String!
}

type B { 
    field: String!
}

union C = A  |  B

type Query { 
    c: C
} 
```

And a query like 

```graphql
query { 
    c  {
        field
    }
}
```

The type of `field` is `String` not `C`

jycouet: I'll add the test to have it covered. 

I added like this because getType() was throwing. There is probably a better way.

I'll look in coming days,unless you have an ideas already?

AlecAivazis (approved):

PR #1107, opened 2023-06-17
Reset Cache Functionality

m4tr1k:
Fixes issue from this Discord [thread](https://discord.com/channels/1024421016405016718/1118674679440363673/1118674679440363673)

### DIFF
--- a/packages/houdini/src/runtime/cache/tests/reset.test.ts
+++ b/packages/houdini/src/runtime/cache/tests/reset.test.ts
@@ -182,7 +182,20 @@ test('make sure the cache subscribers were reset', function () {
 		},
 	}
 
+	// subscribe to the cache
+	const set = vi.fn()
+	cache.subscribe({
+		rootType: 'Query',
+		set,
+		selection,
+	})
+
+	// reset the cache
+	cache.reset()
+
+	// write the data again
 	cache.write({
+		parent: 'User:1',
 		selection,
 		data: {
 			viewer: {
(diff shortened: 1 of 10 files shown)


packages/houdini/src/runtime/cache/tests/reset.test.ts
```diff
@@ -0,0 +1,216 @@
...
+	// subscribe to the cache
+	const set = vi.fn()
+	cache.subscribe({
+		rootType: 'Query',
+		set,
+		selection,
+	})
+
+	// reset the cache
+	cache.reset()
```
AlecAivazis, line 215: I think some of this test is missing. In order to verify that the subscriptions were removed, we just need to call `write` once more and make sure that `set` was never called

AlecAivazis (changes requested):

packages/houdini/src/runtime/cache/tests/reset.test.ts
m4tr1k: @AlecAivazis Thank you for the review. I'm going to work on it!

packages/houdini/src/runtime/public/cache.ts
```diff
@@ -124,4 +124,12 @@ Please acknowledge this by setting acceptImperativeInstability to true in your c
 	): void {
 		return this._internal_unstable.markTypeStale(type ? { ...options, type } : undefined)
 	}
+
+	/**
+	 * Reset the entire cache by clearing all records and lists
+	 */
+
+	reset(): void {
```
AlecAivazis, line 132: 👍

AlecAivazis (approved):
Thanks again!!

PR #1114, opened 2023-06-23
request: interface types

canastro:
Given a Pet union type which as Cat | Dog  (interface Animal\Node) and User (interface Node).
If I have a fragment where I only select Animal interface

```graphql
fragment TestFragment on Pet {
	... on Animal {
		isAnimal
	}
}
```

The current type generated on this fragment is:

```ts
export type TestFragment$data = {} & (({
	readonly isAnimal: boolean;
	readonly __typename: "Cat";
}) | ({
	readonly isAnimal: boolean;
	readonly __typename: "Dog";
}) | ({
	readonly __typename: "non-exhaustive; don't match this";
}));
```

Which means that to access common properties on Animal (which is the interface that I've requested in the fragment) I have to do:

```ts
if (data.__typename === 'Dog' || data.__typename === 'Cat') {
  console.log(data.isAnimal);
}
```

Given my fragment, I would expect to be able to directly do `data.isAnimal`. To be able to do that, I assume that I would need a generated type as:

```ts
export type TestFragment$data = {
	readonly isAnimal: boolean;
} & (({
	readonly isAnimal: boolean;
	readonly __typename: "Cat";
}) | ({
	readonly isAnimal: boolean;
	readonly __typename: "Dog";
}) | ({
	readonly __typename: "non-exhaustive; don't match this";
}));
```

### DIFF
--- a/packages/houdini/src/codegen/generators/typescript/typescript.test.ts
+++ b/packages/houdini/src/codegen/generators/typescript/typescript.test.ts
@@ -21,6 +21,7 @@ const config = testConfig({
 			entity: Entity!
 			listOfLists: [[User]]!
 			node(id: ID!): Node
+			pets: [Pet!]!
 		}
 
 		type Mutation {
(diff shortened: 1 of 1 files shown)


AlecAivazis:
Like I mentioned in discord, Im going to close this so I don't accidentlally split the conversation

PR #1118, opened 2023-06-27
Include types in houdini-svelte/runtime import (resolves QueryStoreFetchParams import failures)

fnimick:
Resolves an issue where `QueryStoreFetchParams` cannot be imported from `$houdini`, meaning that generated operation files do not type-check their variables or event inputs.


AlecAivazis (approved):
Thanks for getting this out so quickly!

PR #1121, opened 2023-06-29
Fix nested fragment arguments

Morstis:
Fixes a bug with nested fragment arguments
If using fragments with nested arguments:
```graphql
fragment QueryFragment on Query @arguments(name: {type: "String!"}) {
    usersByOffset(filter: {name: $name}) {
    id
    }
}
query TestQuery {
  ...QueryFragment @with(name: "Foo")
}
```

houdini will generate this:

```graphql
fragment QueryFragment_32RKor on Query {
  usersByOffset(filter: {name: $name}) {
    id
  }
  __typename
}

query TestQuery {
  ...QueryFragment_32RKor
}
```

=> $name will not be replaced because the graphql.visit()->ArgumentNode() does only search for 1st level argumentNodes.
So it will only find VariableNodes on the first level.

This PR fixes that by deeply exploring the ObjectValuenodes and replacing all the VariableNodes with their explicit values.


AlecAivazis (approved):
Thanks for hunting this down!

AlecAivazis (approved):
🥂

AlecAivazis (approved):

PR #1123, opened 2023-07-03
Bug two subscriptions

Morstis:
Fixes a discord bug where you can not subscribe to two subscriptions from the same page.

A workaround was to use different metadatas. 

```
roleUpdates.listen(undefined, { metadata: { subFix: 1 } }).catch(console.error);
roleDeletes.listen(undefined, { metadata: { subFix: 2 } }).catch(console.error);
```

This fixes it by adding more checks if the subscription requests are the same (the artefact name)
The problem was here: https://github.com/HoudiniGraphql/houdini/blob/main/packages/houdini/src/runtime/client/plugins/subscription.ts#L40 


.changeset/loud-planets-buy.md
```diff
@@ -0,0 +1,5 @@
+---
+'houdini': major
```
AlecAivazis, line 2: This is actually a `patch` change since there isn't any new features or breaking APIs

AlecAivazis (changes requested):

.changeset/loud-planets-buy.md
Morstis: Ah your right. Fixed it

AlecAivazis (dismissed):
Thanks again!

AlecAivazis (approved):

PR #1124, opened 2023-07-04
Fix comment on `@load` directive

SeppahBaws:
I stumbled upon some outdated documentation on the `@load` directive, and updated it to reflect its actual behavior.

Let me know if the updated comment is clear, or if it needs some more clarification.


AlecAivazis (approved):
Thanks!

PR #1127, opened 2023-07-08
Fix when marshal is intentionally an array

Morstis:
Fixes the following bug:

When having an array as scalar type houdini does not marhash the array, because it gets interpreted as a list.
https://github.com/HoudiniGraphql/houdini/blob/7e2977ff5f1e737aebbb606e473708036f303d02/packages/houdini/src/runtime/lib/scalars.ts#L114

E.g.
```ts
houdini.config.js
...
_text: {
  type: 'string[]',

  unmarshal(val) {
    return val;
  },

  marshal(array) {
    console.log(array, typeof array);
  return `{${array.toString()}}`;
  }
}
...
```

would log the values of the array when using this type as an input.

E.g. this mutation would result in "foo" and "bar" as two console.logs with type string

```graphql
mutation updateUser($otherLanguages: _text!) {
  insertUserDataOne(object: {otherLanguages: $otherLanguages}) {
    id
  }
}
```
```ts
.mutate({variables: {otherLanguages: ["foo", "bar"]}});
```

This PR fixes this behaviour for mutations. It currently doesn't look into queries.

A note to the code:
https://github.com/Morstis/houdini/blob/28efb4c35db3e3016e44bf25cab4a858c12ba13f/packages/houdini/src/runtime/lib/scalars.ts#L126
the deep exploration will be slow, because it will explore the whole array. When inputting very big multidimensional arrays this code is not optimal!

### DIFF
--- a/packages/houdini/src/runtime/lib/scalars.ts
+++ b/packages/houdini/src/runtime/lib/scalars.ts
@@ -110,9 +110,12 @@ export function marshalInputs<T>({
 
 			// is the type something that requires marshaling
 			const marshalFn = config.scalars?.[type]?.marshal
+			const scalarType = config.scalars?.[type]?.type
 			if (marshalFn) {
-				// if we are looking at a list of scalars
-				if (Array.isArray(value)) {
+				const shouldBeArray =
+					scalarType?.endsWith('[]') || scalarType?.match(/Array<.*>/) || false
+				// if we are looking at a list of scalars that should not be arrays
+				if (Array.isArray(value) && !shouldBeArray) {
 					return [fieldName, value.map(marshalFn)]
 				}
 				return [fieldName, marshalFn(value)]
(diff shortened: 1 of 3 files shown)


packages/houdini/src/runtime/lib/scalars.ts
```diff
@@ -110,10 +110,27 @@ export function marshalInputs<T>({
...
+					// Count the number of "Array" or "[]" appearances in the scalarType. To Check if the scalarType
+					// should be an array. And then calculate the dimension to check if it is a list of arrays
+					const scalarArrayDim = scalarType!.match(/Array|\[\]/g)?.length ?? 0
+
+					// Return if it is not intentionally an array to skip the value dim calc
+					if (scalarArrayDim === 0) return [fieldName, value.map(marshalFn)]
+
+					// deeply calculate the depth of the array. Explores every entry -> very slow
+					const getArrayDepth = (x: unknown): number => {
+						return Array.isArray(x) ? 1 + Math.max(0, ...x.map(getArrayDepth)) : 0
```
AlecAivazis, line 128: The depth "should be" inferable from the scalar definition, right? If so, I think we should avoid this calculation if we can since a scalar could be used many times in a single response

AlecAivazis (changes requested):

AlecAivazis:
I'm going to close this since its gone a bit stale

PR #1132, opened 2023-07-17
Fix typegen for component queries inside of a route

### LINKED ISSUE #1131: Component query inside of a route folder gets incorrect types
### Describe the bug

(copied over from Discord)

Whenever I have a route, e.g. `/items` in which I have a `+page.gql` and a `MyComponent.svelte`, the types for the `MyComponentQueryVariables` function are generated with the page params instead of the component props.

for example, for a component at the `src/lib/components/MyLibComponent.svelte`, the typings are something like this:
```ts
type MyLibComponentProps = ComponentProps<MyLibComponent>;
export type MyLibComponentQueryVariables = <_Props extends MyLibComponentProps>(args: { props: _Props }) => MyLibComponentQuery$input;
```

however, inside the `src/routes/test/MyRouteComponent.svelte`, the typings are generated like this:
```ts
export type MyRouteComponentQueryVariables = VariableFunction<PageParams, MyRouteComponentQuery$input>;
```
which causes me to not be able to access the props of that component.

It isn't too big of an issue for me, because I only put them in the `routes` folder for testing, before I move them to the `lib` folder, but I keep stumbling across it during testing.

### Severity

annoyance

1. Create a new route with a +page.gql query
2. In that route folder, create a component with a component query
3. In that component query, try adding a `_MyQueryVariables` function of type `MyQueryVariables` (imported from `./$houdini`)
4. See that the `MyQueryVariables` type has the types of the route itself, instead of the component

### Reproduction

https://github.com/SeppahBaws/houdini/commit/447469a1afc749cce7c12df19ffe1f0707794da1

SeppahBaws:
Fixes #1131 

Types for component queries inside of the routes didn't get types generated properly. See issue for full explanation.

### DIFF
--- a/packages/houdini-svelte/src/plugin/codegen/routes/index.ts
+++ b/packages/houdini-svelte/src/plugin/codegen/routes/index.ts
@@ -57,7 +57,10 @@ export default async function svelteKitGenerator(
 			}
 
 			const componentNames: string[] = []
-			const uniqueComponentQueries: { query: OperationDefinitionNode, componentPath: string }[] = []
+			const uniqueComponentQueries: {
+				query: OperationDefinitionNode
+				componentPath: string
+			}[] = []
 			for (const component of componentQueries) {
 				if (!componentNames.includes(component.query.name!.value)) {
 					componentNames.push(component.query.name!.value)
@@ -93,7 +96,11 @@ export default async function svelteKitGenerator(
 					config,
 					uniqueLayoutQueries
 				)
-				const componentQueryTypeImports = getComponentTypeImports(dirpath, config, uniqueComponentQueries)
+				const componentQueryTypeImports = getComponentTypeImports(
+					dirpath,
+					config,
+					uniqueComponentQueries
+				)
 
 				// Util bools for ensuring no unnecessary types
 				const beforePageLoad = pageExports.includes(houdini_before_load_fn)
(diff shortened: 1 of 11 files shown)


packages/houdini-svelte/src/plugin/codegen/routes/index.ts
```diff
@@ -318,6 +374,29 @@ function append_VariablesFunction(
 		.join('\n')
 }
 
+// Copied and adapted from packages\houdini-svelte\src\plugin\codegen\components\index.ts
```
AlecAivazis, line 377: Could you put a comment on the function that this was adapted from just in case we run into a bug over there. I dont want to forget it could be duplicated here

SeppahBaws: Alright, will do it during lunch today or this evening.

Looking through the code however made me think that perhaps these two files could maybe be merged?
Maybe just have one file that goes through the files in the project, and then depending on what type of file it is it uses a different generator for the types? Could be that I'm missing some resoning behind why it's split up though.
But it would eliminate having to duplicate the component query type generation.

SeppahBaws: If I find some time later on I might merge these files, but I think it might be better to do that in a separate PR instead of adding it to this one.

AlecAivazis (approved):
Thanks for adding the comment!

PR #1133, opened 2023-07-18
Fix cache read and write input param types being generated incorrectly

### LINKED ISSUE #1129: Typescript types of variable inputs for cache read and write functions are incorrect
### Describe the bug

When using the Cache API, the requested variables for the query in `Cache.read` or `Cache.write` correspond to the `$result` type (instead of `$input`), which gives an error when using Typescript

Printscreen of the error:

Screenshot 2023-07-11 at 16 21 38

### Severity

annoyance

1. Make sure you have a query that requires variables to be passed through, and that you are using Typescript
2. Use one of the Cache API methods: `write` or `read`
3. When passing the variables to that function, you'll see a type error

### Reproduction

_No response_

SeppahBaws:
Fixes #1129 

The type definition of QueryInput used QueryValue, which resulted into it becoming `MyQuery$result`, instead of the expected `MyQuery$input`.

I'm not 100% sure how to write a proper test for this, suggestions welcome 🙂 

### DIFF
--- a/packages/houdini/src/runtime/public/types.ts
+++ b/packages/houdini/src/runtime/public/types.ts
@@ -131,6 +131,6 @@ export type QueryInput<List, _Target> = List extends [infer Head, ...infer Rest]
 	? Head extends [infer _Key, infer _Value, infer _Input]
 		? _Key extends _Target
 			? _Input
-			: QueryValue<Rest, _Target>
+			: QueryInput<Rest, _Target>
 		: 'Encountered unknown query.Please make sure your runtime is up to date (ie, `vite dev` or `vite build`).'
 	: 'Encountered unknown query.Please make sure your runtime is up to date (ie, `vite dev` or `vite build`).'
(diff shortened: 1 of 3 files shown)


AlecAivazis:
Thanks for getting this out! To write a test for this, you could just use the API in the e2e tests. we run a linter/typechecker on the project as part of the CI flow so it will catch type errors

SeppahBaws:
Running into a very weird bug here while adding a test, will need some more digging to find the root cause 🧐 
It seems to be grabbing the `$result` type from the query above
image
image

packages/houdini/src/runtime/public/types.ts
```diff
@@ -103,34 +103,61 @@ export type ListType<Def extends CacheTypeDef, Name extends ValidLists<Def>> = P
 // then typescript wasn't responding on every change... Having the duplication makes
 // it very responsive.
 
+/**
```
AlecAivazis, line 106: love this comment!

AlecAivazis (approved):
Nice and easy. Thanks again!

PR #1138, opened 2023-08-03
create & init / svelte & react

jycouet:
## Introduction
The idea is to have 2 options `init` or `create`.

### Init
You have an existing svelte or react project, let's add Houdini inside with: `npx houdini@latest init`. 
We try our best to "just" add Houdini, and with git, you will be able to tune to your needs.

### Create
You want to create a new project with Houdini (Svelte or React), here is the command: `npm create houdini@latest`
It will look like this:
image

---

Note(s): 
- Now when we ask "What's the URL for your api?", if it's not successful, we have multiple attempts to be able (during the process) to add headers.


jycouet:
Netlify is using `pnpm 8.6.7`, that's why build is failing over there.
I'll try to bump it later.

jycouet:
Let's close this PR as it's a mix of too many topics

PR #1164, opened 2023-09-22
Add fragment name to missing fragment arguments error

SeppahBaws:
In case there's more than one fragment in a file, it can get tricky to identify which fragment is giving errors, so I added the name of the fragment to the error message to help figure out which one is causing issues.


AlecAivazis:
Seems totally reasonable. Thanks for the quick PR  👍

AlecAivazis (approved):

PR #1203, opened 2023-10-02
localSchema (WriteFile & Watch)

jycouet:
https://github.com/HoudiniGraphql/houdini/assets/5312607/e19b331b-c4a4-4f37-ae54-f96e75f7a5dc


packages/houdini/src/vite/index.ts
```diff
@@ -27,14 +27,25 @@ export default function (opts?: PluginConfig): Plugin[] {
...
+					const config = await getConfig(opts)
 
 					// we need to watch some specific files
-					const schemaPath = path.join(path.dirname(config.filepath), config.schemaPath!)
-					if (minimatch(filepath, schemaPath)) {
-						// if it's a schema change, let's reload the config
-						config = await getConfig({ ...opts, forceReload: true })
-						return true
+					if (config.localSchema) {
+						if (filepath.includes('+schema')) {
```
AlecAivazis, line 34: this would only work if the `+schema` file is the one that changed right? Ideally we could hook into any file that it depends on (so HMR still works if I split my schema up into multiple files). I wonder if there's a way to do that with vite 🤔

jycouet: Ho good point
If you start composing your schema in different files, you will need to restart the server.
And you can do it in at least `js`, `ts`, `gql`, `graphql` files...

In the build step, it's possible to know on what it depend directly, but then you need to build the complete tree!
So on start, building the tree, and as soon as one of the file is touched, retrigger the new tree +reload the schema?

It seems to be a lot, but also quite fun! ^^

---

Alternative: `watchSchema` could have a prop "path" or something? With a glob? (Less sexy, but way easier? ^^)

---

All this should maybe come later? or?

jycouet: Just had another idea 👀
Actually, autocomplete works already as it's a vscode plugin.

The issue is that Houdini will throw an error that a field doesn't exist yet for example. 
But at the throwing time we could look at the schema again to make sure it's in sync? 🤔


Pro: we don't need to watch multiple files, could also benefit remote schema 

Con: adding latency

Any takes?

jycouet:
https://github.com/HoudiniGraphql/houdini/assets/5312607/5d062f31-a282-45e6-a2f9-93bab75117d9

Wiith dep tree 🎉
Can add files, and the list of things to watch will grow

AlecAivazis:
nice!!

AlecAivazis (approved):

PR #1233, opened 2023-11-22
Edited docs of store.fetch method

LazerJesus:
Expanded documenation of expected props of `store.fetch`


AlecAivazis (approved):
Thanks for adding this!

PR #1244, opened 2023-12-19
Fix nested fragments getting masked out incorrectly

### LINKED ISSUE #1243: Field parameters passed through fragments with arguments are incorrectly masked out
### Describe the bug

If a field on a type has an argument, which is getting passed through fragments, the Houdini store will cause that field to be `null`, while the data is being correctly fetched from the api.

It seems very specific to one case, if a param is passed from fragment to fragment, with a fragment on the connection:
```gql
query Test {
  usersConnection(snapshot: "test") {
    ...UsersListFragment @with(someParam: true)
  }
}

fragment UsersListFragment on UserConnection @arguments(someParam: { type: "Boolean!" }) {
  edges {
    node {
      ...UserItem @with(someParam: $someParam)
    }
  }
}

fragment UserItem on User @arguments(someParam: { type: "Boolean!" }) {
  id
  name
  testField(someParam: $someParam)
}
```

If the `UserItem` fragment itself is used in the query on `connection > edges > node`, it works as expected.

**Update:**
I have added another example in my reproduction, showcasing that the fragment masking is causing the field to become `null` after it gets applied.

In the `bug/fragment-params-on-field` route this is clearly shown, as the SSR result gives this:
image

But once the client becomes active and fragment masks are applied, the fields become `null`:
image

### Severity

1. clone the repro link
2. run the dev server
3. go to the `bug/fragment-params-on-connection` route
4. see the data passed via the connection fragment as "Test field: null"

### Reproduction

https://github.com/SeppahBaws/houdini/tree/717e8230d8719b8294532c1bc410ef0c2a34356c

SeppahBaws:
fixes #1243

When fragments with arguments are nested, they ended up not having their arguments in the generated artifact of the parent fragment. This caused the fields in the nested fragment consuming those arguments to be masked out, as those fragments were instantiated with an empty arguments object.


SeppahBaws:
@AlecAivazis I think this PR is pretty much ready, however some of the e2e tests seem to still be failing. Could you take a look at this, as I can't seem to be able to get those running on my computer?

AlecAivazis:
No problem! I updated the snapshots, will merge this in once we get the 👍 from github

AlecAivazis (dismissed):

AlecAivazis (approved):

PR #1249, opened 2024-01-03
Optimize cache subscriptions

llJochemll:
Fixes (can make an issue if needed)

Improves the performance of the functions related to cache subscriptions. Profiling with a large table for a work project resulted in roughly 50% less CPU time (the table uses many subscriptions, this made it go from ~500ms to ~220ms). Most of the improvement comes from line 156, replacing the `map()` and `includes()` with `some()`.

### DIFF
--- a/packages/houdini/src/runtime/cache/subscription.ts
+++ b/packages/houdini/src/runtime/cache/subscription.ts
@@ -354,9 +354,7 @@ export class InMemorySubscriptions {
 
 	reset() {
 		// Get all subscriptions that do not start with the rootID
-		const subscribers = [...this.subscribers.entries()].filter(
-			([id]) => !id.startsWith(rootID)
-		)
+		const subscribers = [...this.subscribers.entries()].filter(([id]) => !id.startsWith(rootID))
 
 		// Remove those subcribers from this.subscribers
 		for (const [id, _fields] of subscribers) {
--- a/packages/houdini/src/runtime/cache/subscription.ts
+++ b/packages/houdini/src/runtime/cache/subscription.ts
@@ -22,16 +22,21 @@ export class InMemorySubscriptions {
 		this.cache = cache
 	}
 
-	private subscribers: {
-		[id: string]: { [fieldName: string]: FieldSelection[] }
-	} = {}
-	private referenceCounts: {
-		[id: string]: { [fieldName: string]: Map<SubscriptionSpec['set'], number> }
-	} = {}
+	private subscribers = new Map<
+		string,
+		Map<
+			string,
+			{
+				selections: FieldSelection[]
+				referenceCounts: Map<SubscriptionSpec['set'], number>
+			}
+		>
+	>()
+
 	private keyVersions: { [key: string]: Set<string> } = {}
 
 	activeFields(parent: string): string[] {
-		return Object.keys(this.subscribers[parent] || {})
+		return Object.keys(this.subscribers.get(parent) || {})
 	}
 
 	add({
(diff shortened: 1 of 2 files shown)


AlecAivazis:
Thanks so much for digging through this! It couldn't have been easy...

Ill have some time tonight to review it but I see all of the tests passing so I'm sure it will be a quick one 🚀

packages/houdini/src/runtime/cache/subscription.ts
```diff
@@ -346,18 +354,16 @@ export class InMemorySubscriptions {
 
 	reset() {
 		// Get all subscriptions that do not start with the rootID
-		const subscribers = Object.entries(this.subscribers).filter(
-			([id]) => !id.startsWith(rootID)
-		)
+		const subscribers = [...this.subscribers.entries()].filter(([id]) => !id.startsWith(rootID))
```
AlecAivazis, line 357: Hm, doesn't `.entries()` return a new list? It's probably neglible but it'd be great to avoid the shallow copy if possible.

same with the `.values` below

packages/houdini/src/runtime/cache/subscription.ts
```diff
@@ -153,21 +167,15 @@ export class InMemorySubscriptions {
 		// add this version of the key if we need to
 		this.keyVersions[key].add(key)
 
-		if (!this.subscribers[id][key].map(([{ set }]) => set).includes(spec.set)) {
-			this.subscribers[id][key].push([spec, selection[1]])
+		if (!subscriberField.selections.some(([{ set }]) => set === spec.set)) {
```
AlecAivazis, line 170: Is this change the core of the improvement? I'm curious if you noticed much of an improvement going from `{}` to a `Map`. I'm totally fine to keep it how you have it, just curious

packages/houdini/src/runtime/cache/subscription.ts
llJochemll: So `.entries()` returns a new iterator, not an array. The spread operator than converts the iterator to an array to be able to use filter on it.

So while `Object.entries()` returns an array, `Map.entries()` returns a reference that contains the information on how to build the array (grossly simplified), and the spread operator uses it to then actually do it. In both cases only one array is actually constructed.

packages/houdini/src/runtime/cache/subscription.ts
llJochemll: To answer the `Map` part first, the performance should be 'better', but on a scale that probably isn't going to matter much. Using Map has advantages, but frankly, my main reason is that the Map api is just nicer to use) (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#objects_vs._maps for a comparison)

llJochemll: The actual improvement comes from combining `map()` and `.includes()` into `some()`.

This is because those functions, unlike in let's say C# (IEnumerable) or Java (Stream), operate on arrays, so chaining methods can actually become quite expansive. The reason for that is that `map()`, `filter()` and the like return a new array, so you end up allocating n arrays, with n being the amount of chained functions.

This doesn't matter that much for small arrays, but for larger arrays it starts to add up.

AlecAivazis:
Sorry I disappeared here! I ended up getting covid while traveling home from the holidays and I only just found the energy to look at my computer. 

The changes look good! Once there's a change set I'll approve the PR 👍

llJochemll:
I've added a changeset

AlecAivazis (approved):

PR #1253, opened 2024-01-24
Allow for required arguments on paginated fragments

### LINKED ISSUE #1143: Paginated fragments with required arguments
### Describe the feature

At the moment, when I have a paginated fragment, I can't have required arguments on that fragment. It would be nice to be able to add `@paginate` on those fragments.

This is what I'm trying to do:
```gql
# In my +page.svelte:
query GroupDetails($id: ID!, $search: String!) {
  group(id: $id) {
    name
    ...GroupMemberships @with(search: $search)
  }
}

# In my GroupMemberships.svelte component:
fragment GroupMemberships on Group @arguments(search: { type: "String!" }) {
  members(filterName: $search) @paginate(name: "MyList") {
    edges {
      node {
        id
        name
        avatar
      }
    }
  }
}
```

I don't immediately see why this wouldn't be able to work, as the fragment should already have the correct values for the required arguments and should simply be able to pass those along with the query when `loadNextPage()` is called.

### Criticality

SeppahBaws:
Fixes #1143

This PR adds functionality to allow required arguments inside paginated fragments.

I'll leave this PR as draft while I test it more thoroughly in our app for work, to make sure nothing else accidentally got broken :)

### DIFF
--- a/packages/houdini/src/codegen/generators/artifacts/inputs.ts
+++ b/packages/houdini/src/codegen/generators/artifacts/inputs.ts
@@ -27,6 +27,12 @@ export function inputObject(
 			}
 		}, {}),
 		types: {},
+		defaults: inputs.reduce((fields, input) => {
+			return {
+				...fields,
+				[input.variable.name.value]: parseInputField(input.defaultValue),
+			}
+		}, {}),
 	}
 
 	// walk through every type referenced and add it to the list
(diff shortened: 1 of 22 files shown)


packages/houdini/src/codegen/generators/artifacts/inputs.ts
```diff
@@ -79,3 +85,33 @@ function walkInputs(
 		{}
 	)
 }
+
+function parseInputField(field: graphql.ValueNode | undefined): any {
```
AlecAivazis, line 89: I had a feeling there was already a similar function in the codebase and it turns out it shows up 3 times 😅  Mind taking the opportunity to consolidate these?

https://github.com/HoudiniGraphql/houdini/blob/main/packages/houdini/src/codegen/generators/artifacts/utils.ts#L43
https://github.com/HoudiniGraphql/houdini/blob/main/packages/houdini/src/runtime/cache/cache.ts#L1505
https://github.com/HoudiniGraphql/houdini/blob/main/packages/houdini/src/runtime/lib/scalars.ts#L139

It seems like 2 of these are defined in the runtime so the one version  _has_ to live there but we can import it into the codegen logic

packages/houdini/src/codegen/generators/artifacts/tests/artifacts.test.ts
```diff
@@ -6672,3 +6692,295 @@ test('nested abstract fragments', async function () {
...
+
+			    "policy": "CacheOrNetwork",
+			    "partial": false
+			};
+
+			"HoudiniHash=f8edbf4199a63a56d214cd5c845d90310052da75c45f3f3f5abf5f5cdb707a3e";
+		`)
+	})
+
+	test('handles complext default arguments', async function () {
```
AlecAivazis, line 6898: typo:

```suggestion
	test('handles complex default arguments', async function () {
```

AlecAivazis (changes requested):
The overall approach looks good! Thanks again

AlecAivazis (approved):
Looks good!

PR #1263, opened 2024-02-20
Ensure houdini generate exits with 1 in case of errors

SeppahBaws:
Fixes https://discord.com/channels/1024421016405016718/1209467376224182312


AlecAivazis (approved):
Thanks for getting this out so quickly!!

PR #1265, opened 2024-02-20
Implement cache policy "NoCache"

### LINKED ISSUE #356: Provide API for fetching data without caching the result
In situations like search, the value doesn't always need to be cached. We should give users a way to fetch, paginate, etc. without writing the values to cache

MattiasMartens:
Fixes #356

Implement the policy "NoCache", mirroring Apollo, which serves a request without reading from or writing to the cache.

Tests are failing for me locally with messages like this:

```
TypeError: Cannot read properties of undefined (reading 'on')
 ❯ vitest.setup.ts:12:1
     10| process.env.HOUDINI_TEST = 'true'
     11| 
     12| beforeEach(clearMock)
       | ^
     13| 
     14| const config = testConfig()
```

This happens even without my changes, but it means I'm not able to write tests for the change.

~~- [ ] If applicable, add a test that would fail without this fix
~~- [ ] Make sure the unit and integration tests pass locally with `pnpm run tests` and `cd integration && pnpm run tests`
- [X ] Includes a changeset if your fix affects the user with `pnpm changeset`

### DIFF
--- a/packages/houdini/src/codegen/transforms/schema.ts
+++ b/packages/houdini/src/codegen/transforms/schema.ts
@@ -18,6 +18,7 @@ enum CachePolicy {
 	${CachePolicy.CacheOnly}
 	${CachePolicy.CacheOrNetwork}
 	${CachePolicy.NetworkOnly}
+	${CachePolicy.NoCache}
 }
 
 """
--- a/packages/houdini/src/runtime/client/documentStore.ts
+++ b/packages/houdini/src/runtime/client/documentStore.ts
@@ -250,7 +250,7 @@ export class DocumentStore<
 				return new Response(JSON.stringify(result.length === 1 ? result[0] : result))
 			}
 
-			// we dont have a proxy so just use the default fetch
+			// we don't have a proxy so just use the default fetch
 			return await globalThis.fetch(input, init)
 		}
 	}
@@ -362,7 +362,7 @@ export class DocumentStore<
 				// invoke the target with the correct handlers
 				const result = target(draft, handlers)
 
-				// if we got _something_ back its a promise so we need to make
+				// if we got _something_ back it's a promise so we need to make
 				// sure something is listening for error
 				result?.catch((err) => {
 					this.#step('error', { ...ctx, index: index - 1 }, err)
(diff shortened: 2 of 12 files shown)


packages/houdini/src/runtime/client/documentStore.ts
```diff
@@ -200,6 +200,8 @@ export class DocumentStore<
 		getSession: () => App.Session | null | undefined
 	): (input: RequestInfo | URL, init?: RequestInit | undefined) => Promise<Response> {
 		return async (input, init) => {
+			// FINDME whatever calls this is what will decide whether/how to cache it.
```
AlecAivazis, line 203: I'm a bit confused what the FINDME comments are referring to here. the caching logic should be totally confined to the `cache` plugin

AlecAivazis (changes requested):
Changes look mostly good! Should be able to merge this in soon.

packages/houdini/src/runtime/client/documentStore.ts
MattiasMartens: Oh no! Trivial mistake on my part. Those were notes I left as I navigated the code. I will remove them.

MattiasMartens:
@AlecAivazis Great! Test snapshots are now updated.

packages/houdini/src/codegen/transforms/schema.ts
```diff
@@ -18,6 +18,7 @@ enum CachePolicy {
 	${CachePolicy.CacheOnly}
 	${CachePolicy.CacheOrNetwork}
 	${CachePolicy.NetworkOnly}
+  ${CachePolicy.NoCache}
```
SeppahBaws, line 21: pretty sure this line will fail the formatting check, due to mismatching indentations
`pnpm format` should do the trick 🙂

SeppahBaws (commented):
Should we maybe also add a unit test to ensure that we're 100% certain it doesn't write to the cache?

MattiasMartens:
@SeppahBaws As I said above, I am not able to write or run tests for this branch! I get this error for most tests:

```
TypeError: Cannot read properties of undefined (reading 'on')
 ❯ vitest.setup.ts:12:1
     10| process.env.HOUDINI_TEST = 'true'
     11| 
     12| beforeEach(clearMock)
       | ^
     13| 
     14| const config = testConfig()
```

SeppahBaws:
@MattiasMartens yes of course, was more of a question in general :)
I can take a look at writing a test when I have some spare time, or maybe @AlecAivazis can if he can find the time before I do.

AlecAivazis:
I had a few moments tonight so I pushed up a test 👍

AlecAivazis (approved):
Thanks for the contribution!

PR #1267, opened 2024-02-27
Non string delete

AlecAivazis:
Fixes n/a

This PR adds support for using `@Foo_delete` on non-string values


knd775:
I've tested this in our project that was previously unable to use `@Item_delete`, and it works in all cases I've been able to test

AlecAivazis:
Thanks for confirming @knd775!

PR #1271, opened 2024-03-01
remove generics from script tag before calling svelte preprocessor

sastan:
Otherwise it will fail to parse the source if the generics attribute contains angle brackets.

For example this code failed to parse:

```html
<script lang="ts" generics="T extends Record<string, unknown>">

</script>
```

Now the `parseSvelte` function will remove the generics attribute before calling the svelte preprocessor
preserving the token positions in the source code.

The output for the above example will be:

```html
<script lang="ts"                                             >

</script>
```

### DIFF
--- a/.changeset/perfect-bags-occur.md
+++ b/.changeset/perfect-bags-occur.md
@@ -1,0 +1,26 @@
+---
+'houdini-svelte': patch
+---
+
+Remove generics from script tag before calling svelte preprocessor.
+
+Otherwise it will fail to parse the source if the generics attribute contains angle brackets.
+
+For example this code failed to parse:
+
+```html
+<script lang="ts" generics="T extends Record<string, unknown>">
+
+</script>
+```
+
+Now the `parseSvelte` function will remove the generics attribute before calling the svelte preprocessor
+preserving the token positions in the source code.
+
+The output for the above example will be:
+
+```html
+<script lang="ts"                                             >
+
+</script>
+```
--- a/packages/houdini-svelte/src/plugin/extract.test.ts
+++ b/packages/houdini-svelte/src/plugin/extract.test.ts
@@ -54,6 +54,57 @@ describe('parser tests', () => {
 		`)
 	})
 
+	test('happy path - typescript with generics', async () => {
+		const doc = `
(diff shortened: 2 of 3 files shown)


SeppahBaws:
Thanks for taking the time to look at this, nice catch!

However, I noticed that the houdini-svelte package is still using svelte version 3.x, and since the `generics=""` syntax is new since svelte 4.x, I think this is the root cause of the issue.

Imho that would also be a nicer fix than a regex [which can sometimes have potential unintended side-effects](https://blog.cloudflare.com/details-of-the-cloudflare-outage-on-july-2-2019/)

sastan:
I thought updating or requiring svelte 4.x would be a breaking change that would/might need a new major version. 

Svelte itself is using a regexp to detect the block: https://github.com/sveltejs/svelte/blob/74f8e261c6da4f2c882ecf76bd34c8e7cce2656b/packages/svelte/src/compiler/preprocess/index.js#L252

I think this is the least invasive option... we could a reminder to remove once houdini-svelte requires svelte 4.x

AlecAivazis:
Yea this is a very unforunate reality of major version bumps like this... I'm not sure if there is a better option here 🤔

sastan:
Without this patch we are unable to start our project as soon as we are using generics in **any** svelte component. We are currently applying this patch ourself — so no hurry. But it would be nice to have this issue fixed in the package.

Regarding the regexp: I really do not see the problem because svelte is doing the same. And that is done is removing the generics attribute and replacing in with space just like the preprocessor a few line late is doing it: https://github.com/HoudiniGraphql/houdini/blob/22d5fd570dab2e9c0a10b2eda27b374e85a6de40/packages/houdini-svelte/src/plugin/extract.ts#L56

SeppahBaws:
Honestly I'm fine with having it in here as a temporary fix, to unblock anyone who is having the same issue.
I have a feeling that once svelte 5 is out and we're adding support for it, we'll have to touch this file again anyways. 😅 

Maybe a comment stating that it's a temporary fix and a link to this PR might be a good idea though. 🙂

SeppahBaws (approved):
Thanks again for the fix!

PR #1277, opened 2024-03-04
Add JSDoc comments generation in $result types and enum declarations

### LINKED ISSUE #1274: Use GraphQL field descriptions to generate JSDoc comments on `$result` types
### Describe the feature

In short, adding JS Doc-style comments on fields in `$houdini/artifacts/QueryName.d.ts` would allow showing description of fields in the IDE.

For example, with the following GraphQL schema:

```graphql
type Query {
    """Some documentation here"""
    example: String!
}
```

and the following document:

```graphql
query Example {
    example
}
```

The following store type definition would be created:

```diff
  export type Example$result = {
+ /** Some documentation here */
      readonly example: string;
  }
```

And would allow to see the documentation of the query on hover in (most?) IDEs.

image

Of course, this would also apply to sub-fields, not just the root `Query` type's fields

### Criticality

gwennlbh:
Fixes #1274

It also adds `@deprecated` jsdoc tags when a field has a `deprecationReason`.

I'm not sure about how to write tests. I added a new page under e2e/kit and updated some tests, but only the generation of jsdoc comments on enum declarations is tested.

The new +page.svelte demonstrates that this works!

image

All tests still pass (on my machine at least)


ewen-lbh:
I could also generate jsdoc on $input types when I think about it. And I guess adding a couple snapshot tests wouldn't hurt. Leaving this in draft until both of those tasks are done

ewen-lbh:
I added a snapshot test for jsdoc generation on artifacts types. All unit tests pass. e2e test fail locally but I suspect that I'm missing some setup (I tried `cd e2e/kit && pnpm run tests`, since `integration/` doesn't exist). Testing manually in e2e/kit works anyway, and end-to-end tests pass in CI so I guess it's okay.

ewen-lbh:
Last thing missing is maybe some documentation. I don't really know where to mention it in the docs though.

.changeset/quick-experts-rest.md
```diff
@@ -0,0 +1,5 @@
+---
+'houdini': minor
```
SeppahBaws, line 2: We're not 100% following the semantic versioning - we usually add new features in the patch versions

```suggestion
'houdini': patch
```

e2e/kit/src/routes/description/+page.gql
SeppahBaws, line 1: I think this e2e test can be removed, since it's being tested in the `houdini` package itself already

SeppahBaws (changes requested):
Looks good overall, just some small notes 👍

SeppahBaws (approved):
Awesome, thanks for the contribution!

PR #1279, opened 2024-03-05
Fix vitePreprocess import in init script

SeppahBaws:
Fixes https://discord.com/channels/1024421016405016718/1214698361106472981

The old import was no longer supported since SvelteKit 2.

### DIFF
--- a/.changeset/big-pianos-pull.md
+++ b/.changeset/big-pianos-pull.md
@@ -1,0 +1,5 @@
+---
+'houdini': patch
+---
+
+Fix init script to use the correct vitePreprocess import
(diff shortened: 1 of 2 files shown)


AlecAivazis:
Thanks for updating this! Is this something we can merge without bumping our sveltekit version?

SeppahBaws:
> Thanks for updating this! Is this something we can merge without bumping our sveltekit version?

I'm unsure. We've been using SvelteKit 2 alongside Houdini for a while now at work without problems so at least the runtime is unaffected.

We could add a check to output the correct svelte config file depending on which sveltekit version the user is using, but I assume that the majority of people starting a project now with svelte and houdini are likely going to be using SvelteKit 2 anyways.

A different route we could maybe explore for more long-term is to simply add the alias key to the existing svelte config, that way it's easier to add Houdini to a project should people decide to migrate from another library.

AlecAivazis (approved):
Ah yea right, the project gets initialized with a package.json that refers to sveltekit 2 so we should be in the clear 👍

PR #1289, opened 2024-03-20
Fix mutation list operations not working when passing `@with` directive

### LINKED ISSUE #1288: List insert mutation with a `@with` param doesn't update the list
### Describe the bug

A mutation with a `...MyList_insert @with(param: value)` spread doesn't update the list in cache, and needs a page refresh for the result to show up.

### Reproduction

https://github.com/HoudiniGraphql/houdini/commit/32e2a5e2beb3f9c84a98868a1616f0772a7a0d9b

SeppahBaws:
Fixes #1288

In the case of doing a `...MyList_insert @with(some: "param")`, the artifact generator that detects mutation list operations ran after the fragment spread hashing step, causing
`...MyList_insert @with(some: "param")`
to be transformed into
`...MyList_insert_asdf`

The code that tries to detect list operations works by checking if the fragment spread name ends with `_insert` or the like. Since we have the hash here at the end, this code would result in the code thinking this was not a list operation fragment.

### DIFF
--- a/.changeset/rotten-cobras-grin.md
+++ b/.changeset/rotten-cobras-grin.md
@@ -1,0 +1,5 @@
+---
+'houdini': patch
+---
+
+Mutation list operations now work if you need to pass a `@with` directive to the fragment spread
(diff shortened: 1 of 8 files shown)


AlecAivazis (approved):
Nice and simple. Thanks for looking into this!

PR #1296, opened 2024-04-13
Fix memory leak in error path

### LINKED ISSUE #1294: Memory leak in error path
### Describe the bug

Hello everyone !

Thank you for this library ! We are actively using it at the present moment and I would need your help regarding a memory leak we discovered.

## Description

We are currently facing an issue where the memory consumption of our docker containers are increasing over time. 
We are hosting a sveltekit application and a backend exposing a graphql API. The application uses Houdini as the fetch client. After some investigation, I can say the issue is coming from the Houdini library.
From what I can see, Houdini stores are not removed from memory if an error is thrown right after a fetch call or if the fetch call itself throws an error.

I created a reproduction repo [here](https://github.com/Domino9697/houdini-memory-leak) based on the pokedex :smile: 

Here is the important part where we artifically throw an error after calling fetch that creates the memory leak:

```js
    const { Info } = await load_Info({
        event,
        variables: {
            id: 2
        })

    // This throws an error
    error(500)

    return {
        Info
    }
}
```

## Steps to reproduce

See the repro repo Readme [here](https://github.com/Domino9697/houdini-memory-leak)

## Expected result

The memory usage should not increase over time.
The number of houdini document stores should not increase over time and should be limited to just the current active stores when the snaphsot was taken.

## Actual result

The memory usage increases over time.
The number of houdini document stores increases over time. I could get thousands of stores in a few minutes.

image

## Investigation comments

When looking at why this is happening, I saw that an active subscription is kept in memory if an error is thrown after calling fetch. I looked at the Houdini source code and found that the issue is coming from [here](https://github.com/HoudiniGraphql/houdini/blob/71ec48593c828696342c30afb8cf430a5c5dec56/packages/houdini-svelte/src/runtime/stores/base.ts#L119).
When we call fetch, we call the BaseStore setup function which subscribed to an observer... And I see that we do unsubscribe from this when a unsubscribe call is made to the Houdini store. However, this might never happen especially in the scenario mentioned above.

This is not an issue in the nominal path because the library expects users to return the store in svelte components and then subscribe to it. However, if the fetch itself throws or if a throw happens after calling fetch, the store is not subscribed to and unsubscribe is not called.

To verify this, we can add a `get` call right after fetch is called and look at the results.

```js
	const { Info } = await load_Info({
		event,
		variables: {
			id: 2
		}
	})

	get(Info)

    // This throws an error
	error(500)

	return {
		Info
	}
}
```

By doing this, we don't see the leak anymore.

## Fix for now

On our side we can add a `get` call right after the fetch call to avoid the memory leak. However, I think this should be fixed in the library itself as this could be a serious issue for a lot of users using manual loading or the throwOnError option.

Since we are getting these issues because we throw when a graphql error is returned, we could use the following hack for now:

```js
    const Info = new InfoStore()
    await Info.fetch({
        event,
        variables: {
            id: 2
    }).catch((err) => {
        get(Info)

        throw err
    })


    return {
        Info
    }
}
```

I have no idea if this would break some functionalities but after some testing, it seems to work fine.

### Reproduction

https://github.com/Domino9697/houdini-memory-leak

AlecAivazis:
This PR fixes a memory leak in the cache that was found as part of the investigation into #1294. This PR does not fix the original issue since it requires some additional thinking and there is a work-around

### DIFF
--- a/.changeset/breezy-waves-cover.md
+++ b/.changeset/breezy-waves-cover.md
@@ -2,4 +2,4 @@
 'houdini': patch
 ---
 
-Fix memory leak
+Fix memory leak in cache's reference counting
--- a/.changeset/breezy-waves-cover.md
+++ b/.changeset/breezy-waves-cover.md
@@ -1,0 +1,5 @@
+---
+'houdini': patch
+---
+
+Fix memory leak
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -6,7 +6,7 @@ on:
     push:
         branches:
             - next
-            - plugin-runtime-search
+            - error-memory-leak
 
 env:
     CI: true
--- a/packages/houdini/src/runtime/cache/cache.ts
+++ b/packages/houdini/src/runtime/cache/cache.ts
@@ -106,6 +106,11 @@ export class Cache {
 
 	// register the provided callbacks with the fields specified by the selection
 	subscribe(spec: SubscriptionSpec, variables: {} = {}) {
+		// if the cache is disabled, dont do anything
+		if (this._internal_unstable.disabled) {
+			return
+		}
+
 		// add the subscribers to every field in the specification
 		return this._internal_unstable.subscriptions.add({
 			parent: spec.parentID || rootID,
@@ -334,7 +339,7 @@ export class Cache {
 
 class CacheInternal {
 	// for server-side requests we need to be able to flag the cache as disabled so we dont write to it
-	private _disabled = false
+	disabled = false
 
 	_config?: ConfigFile
 	storage: InMemoryStorage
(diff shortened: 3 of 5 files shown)


packages/houdini-svelte/src/runtime/stores/query.ts
```diff
@@ -202,7 +202,11 @@ This will result in duplicate queries. If you are trying to ensure there is alwa
 				this.loadPending = false
 				params.then?.(val.data)
 			})
-			.catch(() => {})
+			.catch(() => {
+				if (this.subscriberCount === 0) {
```
dominikg, line 206: hmm, if an error happened, did unsubscribe work before or is count at least 1 here?

AlecAivazis: the store only subscribes if the fetch was successful so the count should be 0 if there has only been an error.

that being said, I think i'm going to approach this a bit differently since it doesn't work in all cases

PR #1309, opened 2024-06-23
Replace Server-side events with Server-sent events in documentation

ewen-lbh:
The acronym SSE actually means "Server-**sent** events" (see https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events)


AlecAivazis:
nice catch! thanks

AlecAivazis (approved):

PR #1300, opened 2024-04-29
Svelte 5 Runes support

SeppahBaws:
With [Svelte 5 now being RC](https://svelte.dev/blog/svelte-5-release-candidate) it's about time to finish this up, so that we have Svelte 5 support right when it comes out.

The main breaking changes Svelte 5 brings for us are Runes, which are incompatible with the old `export let` stuff that Houdini was generating for the query loads etc.

Todo list:
- [ ] (ideally) Test in a bigger project and collect feedback
- [x] Make sure that tests broken by breaking changes are patched (https://svelte-5-preview.vercel.app/docs/breaking-changes)
- [x] Add section in Docs on Svelte 5

## Feedback welcome!

If you would like to try this in your project, bump the version in your package.json to the latest `@canary` tag on npm.
Please feel free to give feedback in here on things that maybe are broken or could be done better.

For an example usage of how to use it in runes mode, check out the examples in the docs: https://deploy-preview-1300--houdinigraphql.netlify.app/guides/svelte-5

### Sidenotes

- ~~I'm not entirely sure whether wrapping the store in a `$derived()` rune is the best way to go about it. I think ideally we would be able to define a custom rune `$graphql(...)` or `$fragment(...)` but I doubt that will happen.~~
- ~~Maybe the `graphql()` and `fragment()` functions could return a rune, so that the `$derived()` is hidden from the user. -> users don't need to edit anything about their queries to upgrade. need to experiment whether it would even work...~~

---

### To help everyone out, please make sure your PR does the following:

- [ ] Update the first line to point to the ticket that this PR fixes
- [x] Add a message that clearly describes the fix
- [x] If applicable, add a test that would fail without this fix
- [x] Make sure the unit and integration tests pass locally with `pnpm run tests` and `cd integration && pnpm run tests`
- [x] Includes a changeset if your fix affects the user with `pnpm changeset`

PR #1315, opened 2024-07-04
Bump dependency versions

SeppahBaws:
Bumps dependency versions:
- vite -> v5
- svelte -> v4
- sveltekit -> v2

Something is still broken atm, but I'm not even sure how to fix it but hopefully I will soon™️
`css-tree` is a dependency of Svelte, but somehow after the upgrades it is getting included in the `houdini-plugin-svelte-global-stores` output, which causes this issue.

Ideas on how to fix this are more than welcome 🙃 


SeppahBaws:
Ok I'm going to close this as it would be too much of a hassle to get this working correctly. I think the only fix for the css-tree problem is upgrading to svelte 5, but that's already being solved in #1300

PR #1322, opened 2024-07-08
Handle unmarshalling lists of custom scalars from the cache

### LINKED ISSUE #1321: Unmarshalling lists of scalars not working
### Describe the bug

We have a custom type defined that has `marshal()` and `unmarshal()` methods, this fails when we have an array.
Here's our temporary solution:

```js
scalars: {
  Money: {
    type: 'Money',
    unmarshal(val) {
      // This is the line that I believe shouldn't exist
      if (Array.isArray(val)) {
        return val.map((v) => new Money(parseInt(v)));
      }
      return val ? new Money(parseInt(val)) : new Money(0);
    },
  }
}
```

### Reproduction

_No response_

SeppahBaws:
Fixes #1321

The marshalling step already had an extra check to marshal the individual values if it encountered an array. The unmarshalling step, however, was not invited to the party and just passed the raw values from the cache to the unmarshalling handler in the config.
Now the unmarshalling step has been invited and will pass the individual values of arrays to the proper `unmarshall` handler in the user's config.

### DIFF
--- a/.changeset/dirty-flies-matter.md
+++ b/.changeset/dirty-flies-matter.md
@@ -1,0 +1,5 @@
+---
+'houdini': patch
+---
+
+Make sure lists of custom scalars are unmarshaled item per item instead of as one list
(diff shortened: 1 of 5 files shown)


AlecAivazis (approved):
🤦

PR #1324, opened 2024-07-09
Add configurable timeout to fetching the remote schema

### LINKED ISSUE #1302: Add configurable timeout for pullSchema
### Describe the feature

Currently, pullSchema uses `fetch` without much more, resulting in a 4-minute timeout (in my case) if the remote api is unreachable. This lead to a lot of confusion as to why my dev server wasn't starting, and it would've been neat to have it say "can't connect to your remote API" sooner.

We could do something like this: https://stackoverflow.com/questions/46946380/fetch-api-request-timeout/57888548#57888548

### Criticality

SeppahBaws:
Fixes #1302 

This adds a timeout to `pullSchema`.
In the rare event that the remote API is unreachable (the request is not cancelled immediately, but just never resolves), `pullSchema` will show a warning that the remote API is unreachable, instead of hanging indefinitely before nodejs cancels the fetch automatically. (which can sometimes take minutes!)


AlecAivazis (approved):
nice and simple!

packages/houdini/src/lib/config.ts
```diff
@@ -183,6 +183,13 @@ export class Config {
 		return this.processEnvValues(env, apiURL)
 	}
 
+	schemaFetchTimeout() {
```
AlecAivazis, line 186: Can we make this an attribute to be in line with the rest?

AlecAivazis (changes requested):

AlecAivazis (approved):

PR #1334, opened 2024-08-02
Fix arguments to patched filesystem functions

### LINKED ISSUE #1326: Houdini vite plugin makes the svelte lsp crash
### Describe the bug

Whenever there's a `<style>` tag somewhere in some Svelte code, while using postcss (for tailwindcss) and houdini, the svelte LSP crashes.

After debugging the svelte LSP itself, the crash seemed to come from vite itself, which led me to believe that somewhere in Houdini's vite plugin it does something wrong, which makes the LSP crash with the following error:
```
Preprocessing failed
TypeError [
	.my-custom-class {
		margin-top: 2rem;
	}
]: [postcss] Expected array buffer, or typed array to be returned for the "source" from the "transformSource" function but got null.
    at new NodeError (node:internal/errors:406:5)
    at assertBufferSource (node:internal/modules/esm/translators:84:9)
    at stringify (node:internal/modules/esm/translators:94:3)
    at createCJSModuleWrap (node:internal/modules/esm/translators:219:12)
    at ModuleLoader.commonjsStrategy (node:internal/modules/esm/translators:301:10)
    at callTranslator (node:internal/modules/esm/loader:273:14)
    at ModuleLoader.moduleProvider (node:internal/modules/esm/loader:278:30)
    at async link (node:internal/modules/esm/module_job:76:21) {
  code: '\r\n\t.my-custom-class {\r\n\t\tmargin-top: 2rem;\r\n\t}\r\n',
  loc: { file: undefined, line: undefined, column: NaN },
  __source: 'Style'
}
```

### Reproduction

https://github.com/SeppahBaws/houdini-postcss-svelte-lsp-crash

fnimick:
When calculating a string filepath, use `URL.pathname` rather than `.toString()` if the given object is a URL.

Also pass the given filepath argument through to the underlying fs functions rather than using our generated string filepath.

These resolve an issue where `.toString()` on a URL returns a string prefixed with 'file://' even on POSIX devices, which cannot be used by the nodejs runtime bundled with Electron.

Fixes #1326

### DIFF
--- a/.changeset/eighty-fishes-shake.md
+++ b/.changeset/eighty-fishes-shake.md
@@ -1,0 +1,5 @@
+---
+'houdini-svelte': patch
+---
+
+Fix filesystem patches to not use illegal paths
--- a/packages/houdini-svelte/src/plugin/fsPatch.ts
+++ b/packages/houdini-svelte/src/plugin/fsPatch.ts
@@ -1,8 +1,7 @@
 import type { PluginHooks } from 'houdini'
 import { fs, path } from 'houdini'
-import type { PathLike } from 'node:fs'
-import filesystem, { Dirent } from 'node:fs'
-import filesystemPromises from 'node:fs/promises'
+import filesystem, { Dirent, type PathOrFileDescriptor, type PathLike } from 'node:fs'
+import filesystemPromises, { type FileHandle } from 'node:fs/promises'
 
 import { _config } from '.'
 import type { Framework } from './kit'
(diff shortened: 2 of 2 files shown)


SeppahBaws (commented):
Since we now just end up passing `fp` to the native `readFileSync` functions, do we even need to have the `getStringFilePath` function?

fnimick:
> Since we now just end up passing `fp` to the native `readFileSync` functions, do we even need to have the `getStringFilePath` function?

I'm not sure, but I figure using the correct file path for the patch logic rather than a URL which may very will have query strings, etc can't hurt

SeppahBaws (approved):
Fair enough. Thanks for digging into this!

PR #1345, opened 2024-08-22
Mention include: "credentials" in authentification docs

ewen-lbh:
Hi! I spent way too much time trying to figure out why "Houdini isn't passing cookies to the API while Svelte is", after switching to OAuth2-backed authentication with HTTP-Only cookies.

This is just a friendly reminder, not something Houdini-specific, but it'll surely prevent some tickets ^^


AlecAivazis (approved):
nice and simple! love it.

PR #1349, opened 2024-09-08
Cancel duplicated pagination fetches

SeppahBaws:
When a pagination query takes some time to execute, it is possible to call `loadNextPage` again before the first fetch has finished. This will then give data for the same page  twice, and Houdini simply adds the data to the list.
This can cause some major problems whenever you have a keyed each-loop in svelte, causing a major crash of the application.

This PR adds a check around the `loadNextPage` and cancels the request whenever one for the same page is already in flight.

### DIFF
--- a/.changeset/serious-penguins-unite.md
+++ b/.changeset/serious-penguins-unite.md
@@ -1,0 +1,5 @@
+---
+'houdini': patch
+---
+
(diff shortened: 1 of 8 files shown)


SeppahBaws:
closing this in favor of #1352

PR #1350, opened 2024-09-10
Generate ts config before building

AlecAivazis:
This PR fixes a bug I ran into when running dev on a fresh project that uses a local schema.


SeppahBaws:
Will this overwrite an existing config if the app is being built on the 2nd or nth time?

PR #1352, opened 2024-09-11
Dedupe directive

AlecAivazis:
This PR is an alternative to #1349 in the form of a new directive: `@dedupe(cancelFirst: Boolean)`. The `cancelFirst` argument decides wether the second request is canceled (as in the case of pagination query) or if the first request needs to be canceled (as in the case of trellix's drag and drops) 


packages/houdini/src/codegen/transforms/schema.ts
```diff
@@ -52,6 +52,14 @@ directive @${config.paginateDirective}(${config.listOrPaginateNameArg}: String,
 """
 directive @${config.listPrependDirective} on FRAGMENT_SPREAD
 
+"""
+	@${
+		config.dedupeDirective
+	} is used to prevent an operation from running twice. If the operation is sent
+	while there is already one pending and the cancelFirst arg is true, the first one will be canceled instead of the second.
```
SeppahBaws, line 59: Maybe something along the lines of this is a little bit clearer?

```suggestion
	@${
		config.dedupeDirective
	} is used to prevent an operation from running more than once at the same time.
	If the cancelFirst arg is set to true, the response already in flight will be canceled instead of the second one.
```

packages/houdini/src/runtime/client/documentStore.ts
```diff
@@ -24,6 +24,8 @@ const steps = {
 	backwards: ['end', 'afterNetwork'],
 } as const
 
+let inFlightPageFetches: Record<string, AbortController> = {}
```
SeppahBaws, line 27: Technically these aren't just `pageFetches` anymore now :)

SeppahBaws (commented):
Will check at work tomorrow if functionality still looks ok 🙂

packages/houdini/src/codegen/transforms/schema.ts
AlecAivazis: Yea much better!

packages/houdini/src/runtime/client/documentStore.ts
AlecAivazis: good point haha

packages/houdini/src/codegen/transforms/paginate.test.ts
```diff
@@ -1061,7 +1061,7 @@ test('query with forwards cursor paginate', async function () {
 
 	// load the contents of the file
 	expect(docs[0]?.document).toMatchInlineSnapshot(`
-		query Users($first: Int = 10, $after: String) {
+		query Users($first: Int = 10, $after: String) @dedupe {
```
AlecAivazis, line 1064: @SeppahBaws just to call this out to open the discussion. This is kind of a strange side effect of the current implementation. Since we just reuse the query for a paginated query, i added this directive to the users query even though they didn't specifically do it. This could change existing behavior... I "better" approach would be to generate a pagination query that's a duplicate of the user's query with the extra directive.

SeppahBaws: The only case I can think of that this will break is something like this:

```gql
query Users($filter: UserFilter) {
  users(filter: $filter) {
    edges {
      node{
        id
        name
        age
      }
    }
  }
}
```

```html
<script>
  const query = // the query above.

  let youngUsers = [];
  let oldUsers = [];

  const loadData = () => {
    // fetch both groups in parallel
    query.fetch({variables: {filter: {ageGroup: "young"}}}).then(res => youngUsers = res.data.users.edges);
    query.fetch({variables: {filter: {ageGroup: "old"}}}).then(res => oldUsers = res.data.users.edges);
  }
</script>

<div>
  {youngUsers.map(edge=>edge.node.name).join(", ")}
</div>

<div>
  {oldUsers.map(edge=>edge.node.name).join(", ")}
</div>
```

If we were to `@paginate` the users, only the young users would end up showing up.
But we wouldn't even be able to load the next page, since pagination is done on the query, and not on the result itself.

So I think it's probably fine to just leave it like it is right now...

SeppahBaws (approved):
Been giving this branch a spin in our repo at work, and seems to work well without any problems!

AlecAivazis:
Thanks for confirming @SeppahBaws!

PR #1373, opened 2024-10-21
Add support for Svelte 5

SeppahBaws:
This PR brings Svelte 5 support out of preview mode and will cause some big version bumps:
- houdini: v1.3.0
- houdini-svelte: v2.0.0


AlecAivazis (approved):
🎉

PR #1376, opened 2024-10-23
Fix init script using incorrect houdini-svelte version

SeppahBaws:
When running `npx houdini init` on a brand new svelte project, the init script would incorrectly assume the version for the `houdini` package is the same as the version for the `houdini-svelte` package.
This PR fixes it by altering the build script to fetch the package.json files for all the packages, and then either use `houdini`'s version, or `houdini-svelte`'s version 

### DIFF
--- a/.changeset/thirty-melons-hug.md
+++ b/.changeset/thirty-melons-hug.md
@@ -1,0 +1,5 @@
+---
+'houdini': patch
+---
+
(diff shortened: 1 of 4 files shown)


SeppahBaws:
Needed to tweak the github action to use the new [sv](https://www.npmjs.com/package/sv) cli tool since it replaces create-svelte

AlecAivazis (approved):

PR #1378, opened 2024-10-24
Only dedupe pagination queries if the variables match

AlecAivazis:
This PR adds a new argument to the `@dedupe` directive: `match` which takes 3 values: `Variables`, `Operation`, and `None`. This gives the user more fine-grained control over how the deduping happens. Most importantly, for paginated queries, `@dedupe` comes with `match` set to `Variables`.


SeppahBaws (changes requested):
Hmm this doesn't seem to be doing any deduping at all, regardless of what params I give it...

I'll try to add a small repro in the e2e over the weekend

half2me:
Can we also either have the option to opt in/out of dedupe behaviour? Or at least disable it for SSR?
Currently all my tests are failing when running in parallel mode, because when multiple requests hit my endpoint that uses a `@loading` directive it throws a 500 with `Aborted`.

AlecAivazis:
@half2me it should only be automatically deduping pagination queries. are you seeing something different?

half2me:
> @half2me it should only be automatically deduping pagination queries. are you seeing something different?

Yes I am. Since I upgraded to this, a lot of tests started failing with the server returning status `500` `Aborted` error, for any page where I'm using pagination.

This definitely has something to do with multiple requests hitting the same page, since if I disable parallel workers for my tests, and run them just one at a time, they all work fine. Its only when simultaneous requests hit the same page that they break.,

AlecAivazis:
@SeppahBaws i think i fixed what was wrong

@half2me  i added a new config value `supressPaginationDeduplication` that you can set to prevent the deduplication

half2me:
Amazing, I will rerun the tests and let you know.

SeppahBaws:
Ah awesome, I'll give this a shot once it's calmed down a bit more at work...

@half2me you can clone and build the repo locally, and then link it locally to your repo with `npm/pnpm link`:
https://docs.npmjs.com/cli/v6/commands/npm-link
https://pnpm.io/cli/link

(on npm you can also specify the package version as `"file:../path/to/houdini/packages/houdini"` - I did find that this messes up in pnpm though)

SeppahBaws (approved):
Looking good!

SeppahBaws (approved):

PR #1383, opened 2024-11-05
Update verify_init script to support changed sv CLI options

SeppahBaws:
This fixes currently broken builds, because of the deprecated CLI options in the svelte cli.
I also pinned the version of the cli to the current latest, 0.6, so that we don't have to do this again every time they push a non-backwards compatible change.


AlecAivazis (approved):
Thanks for getting this out

PR #1389, opened 2024-11-11
Take LoadingType object union into account for FragmentStore

### LINKED ISSUE #1388: FragmentStore's _Data type argument does not accomodate @loading fragment's data shape
### Describe the bug

The current definition of `FragmentStore` enforces that `_Data` extends `GraphQLObject`:
```ts
export declare class FragmentStore<_Data extends GraphQLObject, _ReferenceType extends {}, _Input extends GraphQLVariables = GraphQLVariables> {
```

which is defined as follows:

```ts
export type GraphQLObject = {
    [key: string]: GraphQLValue;
};
export type GraphQLDefaultScalar = string | number | boolean;
export type GraphQLValue = GraphQLDefaultScalar | null | GraphQLObject | GraphQLValue[] | undefined;
```

however, `@loading` fragments' `$data` types are defined with a union to represent the loading state, that uses `typeof PendingValue`, but declare that they return `FragmentStore` instances.

This prevents library users from retrieveing the `$data` type given a FragmentStore instance type, somthing which is useufl to extend the library with generic utilities around loading states (i'm working on a Loading class that would allow `.map`-ing obejcts in potentially loading states for much better ergonomics, insteada of having to pass things to utility functions as I currently do)

See further discussion at https://discordapp.com/channels/1024421016405016718/1305501698244214794/1305501698244214794

### Reproduction

https://stackblitz.com/edit/github-t4rqmk?file=src%2Flib%2Findex.ts&view=editor

gwennlbh:
Fixes #1388

By defining new GraphQLLoadedValue & GraphQLLoadedObject types and including LoadingType in GraphQLValue's definition, every type in the library take could accept a loading value reflects this fact in its types. This allows typescript to infer things correctly

### DIFF
--- a/.changeset/sour-starfishes-push.md
+++ b/.changeset/sour-starfishes-push.md
@@ -1,6 +1,7 @@
 ---
 'houdini': minor
 'houdini-svelte': patch
+'houdini-react': patch
 ---
 
 Add new types GraphQLLoadedValue and GraphQLLoadedObject, and include LoadingType in GraphQLValue
--- a/.changeset/sour-starfishes-push.md
+++ b/.changeset/sour-starfishes-push.md
@@ -1,0 +1,6 @@
+---
+'houdini': minor
+'houdini-svelte': patch
+---
+
+Add new types GraphQLLoadedValue and GraphQLLoadedObject, and include LoadingType in GraphQLValue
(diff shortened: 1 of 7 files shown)


ewen-lbh:
okay so this is kinda awful, and I'm down to a `Cache.write` function call - idk how to handle loading values there - but since i'm only changing types, runtime already passes PendingValues where it "shouldn't" so i guess it's fine?

AlecAivazis:
@ewen-lbh any updates here? happy to revisit this when you're ready to see it through

ewen-lbh:
i'm hoping to get back to it this weekend, exams have been piling up cuz it's almost the end of the semester '^^

AlecAivazis:
I'm going to close this for now. feel free to re-open this effort when you have time @ewen-lbh

PR #1392, opened 2024-11-14
Fix documentation parameter type spelling error

damoonrashidi:
This is just an update to a spelling error in the docs. No tests need to run and there is no change in the API that would require generateing a changeset.

These aren't applicable in this case

~- [ ] Update the first line to point to the ticket that this PR fixes~
~- [ ] Add a message that clearly describes the fix~
~- [ ] If applicable, add a test that would fail without this fix~
~- [ ] Make sure the unit and integration tests pass locally with `pnpm run tests` and `cd integration && pnpm run tests`~
~- [ ] Includes a changeset if your fix affects the user with `pnpm changeset`~

### DIFF
--- a/site/src/routes/api/graphql-magic/+page.svx
+++ b/site/src/routes/api/graphql-magic/+page.svx
@@ -60,7 +60,7 @@ mutation UncompleteItem($id: ID!) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


SeppahBaws (approved):
Good catch, thanks for fixing this! 🙂

PR #1398, opened 2024-12-02
Move svelte, sveltekit and vite into peerDependencies

SeppahBaws:
The latest Svelte update added a new element (`<svelte:boundary>`). However, upgrading an existing houdini + svelte project to this latest version gives errors when generating, as the version of the svelte compiler that houdini uses is outdated.

By moving the svelte dependency into peerDependencies, we use the svelte version that the project is using. The only requirement we now have is that it matches `^5.0.0`.

(The updated wrangler version in e2e/react is because I was having compilation errors in better-sqlite3 while installing)


AlecAivazis (dismissed):

knd775:
I think it may be worth considering doing the same with vite

SeppahBaws:
> I think it may be worth considering doing the same with vite

Ah yeah good point. I might probably also move svelte-kit as well...

AlecAivazis (approved):

PR #1400, opened 2024-12-06
Fix credentials key, value order

damoonrashidi:
The property is called `credentials` and the value `include`. This only touches documentation so no tests have been changed.

Fixes #TICKET


AlecAivazis (approved):

AlecAivazis:
Thanks for the quick fix!

PR #1404, opened 2024-12-17
Give better error when receiving non-JSON response

Mtn-View:
Fixes # N/A : https://discord.com/channels/1024421016405016718/1301635915663278131

Without this, if the server returns a non-JSON error response such as an NGINX error page, `defaultFetch` will still attempt to parse the response as JSON, which will throw a syntax error. Instead, this PR changes it to throw the semantic HTTP error which is easier to understand than a SyntaxError from parsing HTML as JSON.

### DIFF
--- a/.changeset/fluffy-lamps-provide.md
+++ b/.changeset/fluffy-lamps-provide.md
@@ -2,4 +2,4 @@
 'houdini': patch
 ---
 
-Throw the semantic HTML error code and message when receiving a non-JSON error response from the server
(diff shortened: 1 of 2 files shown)


SeppahBaws (changes requested):
Maybe it could be useful if the error also contained something like "failed to fetch: server returned invalid response" or so, to be a bit more clear as to what went wrong.

Mtn-View:
Okay, the error message has been updated.

SeppahBaws (approved):
Awesome, thanks for adding this in!

PR #1411, opened 2025-01-13
required check accept fragmentspread

### LINKED ISSUE #1410: Not possible to mark parent as `@required` if parent is nullable due to child being a nullable fragment
### Describe the bug

TLDR; if a query has a fragment with a `@required` in a top level field, it bubbles up to the query, but it's not possible to mark the parent of the fragment spread `@required`.

Example:

I have a fragment
```gql
fragment AirportParts on airports @required {
    id
    location @required {
        city
    }
}
```

And i use it in a query
```gql
query MyQuery {
    flights {
        to {
            ...AirportParts @mask_disable
        }
    }
}
```

Naturally, it leads to a nullable `to`:
```ts
export type MyQuery$result = {
    /**
     * An array relationship
    */
    readonly flights: ({
        /**
         * An object relationship
        */
        readonly to: { // <----- to IS NULLABLE
            readonly id: number;
            readonly iata: string | null;
            /**
             * An object relationship
            */
            readonly location: {
                readonly city: string;
                readonly name: string;
            };
            readonly " $fragments": {
                AirportParts: {};
            };
        } | null; // <----- to IS NULLABLE
    })[];
};
```

However, it is **not possible** to mark the `to` as required:

If you try:
```diff
query MyQuery {
    flights {
-        to {
+        to @required {
            ...AirportParts @mask_disable
        }
    }
}
```

```
❌ Encountered error in src/routes/.../query.gql
@required may only be used on nullable fields
```

### Reproduction

_No response_

joleeee:
Fixes #1410

### DIFF
--- a/packages/houdini/src/codegen/validators/typeCheck.ts
+++ b/packages/houdini/src/codegen/validators/typeCheck.ts
@@ -373,16 +373,24 @@ export default async function typeCheck(config: Config, docs: Document[]): Promi
 }
 
 function validateRequiredDirective(config: Config, filepath: string) {
-	function isClientNullable(node: graphql.FieldNode, ignoreCurrent: boolean): boolean {
+	function isClientNullable(node: graphql.FieldNode | graphql.FragmentSpreadNode, ignoreCurrent: boolean): boolean {
 		const hasRequiredDirective = node.directives?.some(
 			({ name }) => name.value === config.requiredDirective
-		)
-		return (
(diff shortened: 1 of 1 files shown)


AlecAivazis:
Thanks for putting this together @joleeee! I think the change makes sense so once you've added a changeset and a test [here](https://github.com/HoudiniGraphql/houdini/blob/main/packages/houdini/src/codegen/validators/typeCheck.test.ts) it should be good to go.

PR #1412, opened 2025-01-14
Ensure session data is always returned

### LINKED ISSUE #1405: Returning in top-level +layout.ts causes blank session data
### Describe the bug

_I think this is a bug?_ It's certainly weird behavior. If you have a return statement in `src/routes/+layout.[ts,js]`, any value you set with `setSession` in `hooks.server.js` will be lost. 

### Reproduction Steps

1. Load up stack blitz below
2. Wait for deps to install and homepage to load.
3. Observe `fetchParams session {}` in console server log.
4. Open `src/routes/layout.js` and remove `if (!browser) return;`, or change it to 
```js
  if (browser){
    console.log('this should only run in the browser')
  }
```
5. Press refresh on homepage. Observe server console log now contains the token defined in `src/hooks.server.js`
```
fetchParams session { token: 'asdf' }
```

This doesn't appear to happen with nested `+layout.js` files, only the top leve, though my testing was limited outside of trying to reproduce with as little code as possible.

> Note: The stackblitz displays the "result values" to screen, which are the contents of `ctx.sesion`. When the page first loads, you can see `data` is null for a moment, then it updates on the client to a valid value. I'm not really sure why it's "double fetching" on the frontend and backend, I couldn't figure out how to make it SSR only.

The key components of the stackblitz:

- `src/client.js:fakeFetchPlugin` A fake plugin to mimic a network data return. 
    - This plugin has a console.log/console.warn, depending on if `ctx.session.token` is defined. 
    - This plugin returns the value of `ctx.session` as the returned data.
    - This plugin has a `throw new Error` that can be used to mimic what would really happen if you tried to make an unauthenticated network request where authentication was required.
- `src/hooks.server.js` - This uses `setSession` from houdini so set a static session value, mocking the process of setting a real token.
- `+layout.js` is the problem file, and just returns early if not running in the browser. This mimics what I was doing on a real project when I discovered this "bug".
- The rest of the code is basic boilerplate just to get things running. I kept it as slim as possible.

### Reproduction

https://stackblitz.com/edit/svelte-kit-3o9ngm2q?file=package.json

AlecAivazis:
Fixes #1405 

This PR fixes an issue with the transformation of the root layout if it included an early return. Previously the logic would look for a single value returned but now we look for _any_ return in the function and ensure that the necessary values are included

### DIFF
--- a/.changeset/large-lamps-complain.md
+++ b/.changeset/large-lamps-complain.md
@@ -1,0 +1,5 @@
+---
+'houdini-svelte': patch
+---
+
+Fix issue with sessions if there are early returns in root layout load
--- a/package.json
+++ b/package.json
@@ -62,9 +62,5 @@
 		"overrides": {
 			"graphql": "15.5.0"
 		}
-	},
-	"engines": {
-		"node": "^20",
-		"pnpm": "^8"
 	}
 }
--- a/packages/houdini-svelte/src/plugin/transforms/kit/session.test.ts
+++ b/packages/houdini-svelte/src/plugin/transforms/kit/session.test.ts
@@ -318,3 +318,20 @@ test('passes session from root client-side layout', async function () {
 		}
 	`)
 })
+
+
+test('augments load function in root layout load', async function () {
+	const result = await test_transform_js('src/routes/+layout.js', `
+		import { browser } from '$app/environment';
+
+		export const load = async ({ url }) => {
+			console.log('routes/+layout.js start');
+			if (!browser) return;
+
+			console.log('this should only run in the browser');
+		};
+	
+	`)
+
+	expect(result).toMatchInlineSnapshot(``)
+})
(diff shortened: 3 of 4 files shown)


Rudedog9d:
@AlecAivazis I was able to successfully test this against the [test project](https://stackblitz.com/edit/svelte-kit-3o9ngm2q?file=package.json) from the original issue.

Steps for testing were:

```bash
# clone and setup houdini
git clone git@github.com:HoudiniGraphql/houdini.git
pnpm i

# build on main to replicate original issue
pnpm build

# install local houdini-svelte in test project
npm i ../houdini/packages/houdini-svelte/

# verify the server outputs "fetchParams session {}"
npm run dev

# checkout this branch
git checkout root-layout-load

# rebuild houdini-svelte - I used some extra options here because I was getting some issues with old code being cached
pnpm i && pnpm build --no-cache --force

# re-install local houdini-svelte in test project
npm i ../houdini/packages/houdini-svelte/

# verify the server outputs "fetchParams session { token: 'asdf' }"
npm run dev
```

I was able to successfully switch between main and this branch with expected results several times using this method :+1:

PR #1430, opened 2025-01-27
Fix peerdependencies versions

SeppahBaws:
This PR bumps the peerdependency versions of vite and sveltekit, as we apparently seem to have missed these between the peerdependencies and vite 6 PRs.

### DIFF
--- a/packages/houdini/package.json
+++ b/packages/houdini/package.json
@@ -60,7 +60,7 @@
         "vite-plugin-watch-and-run": "^1.7.0"
     },
     "peerDependencies": {
-        "vite": "^6.0.3"
+        "vite": "^5.3.3 || ^6.0.3"
     },
     "files": [
         "build"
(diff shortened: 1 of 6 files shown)


packages/houdini/package.json
```diff
@@ -61,7 +60,7 @@
         "vite-plugin-watch-and-run": "^1.7.0"
     },
     "peerDependencies": {
-        "vite": "^5.3.3"
+        "vite": "^6.0.3"
```
dominikg, line 63: i think this would be a breaking change. if you want to add vite6, that can be done in a minor, but removing vite5 would be breaking. (a user stuck on vite5 cannot update houdini anymore)

SeppahBaws: Hmm that's fair. So something like this would be a better solution then?
```suggestion
        "vite": "^5.3.3 || ^6.0.3"
```

SeppahBaws:
Alright everything should be good now.
Tested it with our work project and should still be backwards compatible with projects running on vite 5.

AlecAivazis (approved):

PR #1455, opened 2025-03-22
feat(HMR)!: integrate Vite 6 HMR -> only run code generation pipeline if modfied file has graphql dependency, reduce HMR event count

rgon:
Fixes #1454

This PR refactors the dev server file watcher so it runs from standard Vite HMR events. It then only runs the houdini code generation pipeline if the file imports from the `$houdini` (or `config.runtimeDir`) path. It finally only asks for module replacement if there are changes in the generated code.

It reduces the amount of HMR events on any change in a project which imports Houdini from ~100 to 1, being less of a toll on the client browser.

It speeds up HMR events on the project: only slowing down those that really have graphql code niside.

This will set an easier base for fully atomic rebuilds.
 
---

#### Current build pipeline on dev server
On every change, the `houdini/src/codegen/index.ts compile` method is called from `houdini/src/vite/index.ts watchAndRun`.
https://github.com/HoudiniGraphql/houdini/blob/aff81e3698099270c2cccd315b2679f3061564ae/packages/houdini/src/vite/index.ts#L29C1-L33C40

```js
async run(server, absolutePath:string|null)
```

#### Proposed alternative
This `watchAndRun` method appears not to be necessary. Vite provides the [`hotUpdate` handler](https://vite.dev/guide/api-environment-plugins.html#the-hotupdate-hook) for create/update/delete, with a list of modules (useful later!) which are used to detect whether the file change should trigger a houdini rebuild or not.

https://github.com/HoudiniGraphql/houdini/blob/aff81e3698099270c2cccd315b2679f3061564ae/packages/houdini/src/vite/houdini.ts#L25

We just need to move the contents of the `watchFile` and `run` methods to this new handler.

#### Potential issues
Since HMR detection is handled separately from actual builds (which are on `buildStart`), this change probably does no damage. However, if there were a way to test HMR builds programatically, it would be smart. This is my first dive into both vite and houdini, so I need a hand on this!

## Task list/PR Status
+ [x] Find which function triggers the rebuild, and who rebuilds any file.
+ [x] Implement `hotUpdate` alternative ways of listening to file changes
+ [x] Cleanup: if working, deprecate watchAndRun and their dependencies
+ [x] Trigger HMR replacement only if something has changed
+ [x] Invalidate & successfully trigger update on houdini files
+ [x] detect if the file has a houdini dependency before calling build

PR #1466, opened 2025-04-02
deps(e2e tests): upgrade vite from v4 to v6, since v4 is not actually supported …

rgon:
Required for #1455 (hot update API)

The `e2e/react` and `e2e/svelte` project have an incompatible vite version (4.x.x), they should be 5 or 6.

https://github.com/HoudiniGraphql/houdini/blob/64fe7964266773f0b3eda4ee779481c811b6bfae/packages/houdini/package.json#L63

No breaking changes, so it's a plain dependency update:
V4-V5 https://v5.vite.dev/guide/migration.html
V5-V6 https://vite.dev/guide/migration.html

No changeset needed

### DIFF
--- a/e2e/react/package.json
+++ b/e2e/react/package.json
@@ -48,7 +48,7 @@
 		"postcss": "^8.4.31",
 		"tailwindcss": "^3.3.3",
 		"typescript": "^4.9.3",
-		"vite": "^4.1.0",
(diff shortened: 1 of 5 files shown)


AlecAivazis (approved):
Thanks for putting this together!

PR #1469, opened 2025-04-11
Fix @load directive not using renamed props

### LINKED ISSUE #1468: Svelte 5: Component queries with `@load` don't work with renamed props
### Describe the bug

The code here doesn't take renamed props into account.
https://github.com/HoudiniGraphql/houdini/blob/58c66c06a948f0b73eb315150a414d09abab1f21/packages/houdini-svelte/src/plugin/transforms/componentQuery.ts#L46-L74

See example below: the `class` prop has been renamed to `className` to avoid clasing with the js keyword, but Houdini hasn't picked up that this prop was renamed, and tries to generate code with the old prop name.

```svelte
<script lang="ts">
    import { graphql } from "$houdini";
    import type { CharacterDetailsVariables } from "./$houdini";

    interface Props {
        class?: string;
        characterId: string;
    }
    let { class: className = "", characterId }: Props = $props();

    export const _CharacterDetailsVariables: CharacterDetailsVariables = ({ props }: { props: Props }) => {
        return {
            characterId: props.characterId,
        };
    };

    let store = $derived(
        graphql(`
            query CharacterDetails($characterId: ID!) @load {
                character(id: $characterId) {
                    name
                }
            }
        `)
    );
</script>

<div class={className}>
    {#if $store.data}
        <p>{$store.data.character?.name}</p>
    {/if}
</div>
```

```
12:03:33 [vite] (ssr) Error when evaluating SSR module /src/routes/+page.svelte: src/routes/TestComponent.svelte:42:32 Unexpected token
https://svelte.dev/e/js_parse_error
  Plugin: vite-plugin-svelte
  File: src/routes/TestComponent.svelte:42:32
   40 |              input: _CharacterDetailsVariables.call(new RequestContext(), {
   41 |                  props: {
   42 |                      class: class,
                                         ^
   43 |                      characterId: characterId
   44 |                  }
[Error [CompileError]: src/routes/TestComponent.svelte:42:32 Unexpected token
```

### Reproduction

_No response_

SeppahBaws:
Fixes #1468

The generated load functions for component queries were not taking renamed props into account in Svelte 5, and ignoring them completely in legacy svelte mode. This PR fixes this bug.

### DIFF
--- a/.changeset/orange-apricots-kiss.md
+++ b/.changeset/orange-apricots-kiss.md
@@ -1,0 +1,5 @@
+---
+'houdini-svelte': patch
+---
+
+Properly generate load function with renamed props for component queries with an `@load` directive
--- a/packages/houdini-svelte/src/plugin/transforms/componentQuery.test.ts
+++ b/packages/houdini-svelte/src/plugin/transforms/componentQuery.test.ts
@@ -477,7 +477,12 @@ describe('Svelte 5 runes', function () {
 					}
 				}
 	
-				const { prop1: renamedProp } = $props();
+				const {
+                    prop1: renamedProp,
+                    prop2: alsoRenamed = "with default",
+                    prop3,
+                    prop4 = "some default"
+                } = $props();
 	
 				const result = $derived(
 					graphql\`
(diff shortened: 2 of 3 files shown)


SeppahBaws:
Ok seems like it does not _quite_ work perfectly, as it causes some weird behaviour with `let {class: className}` - investigating...

SeppahBaws:
Aha, should be good to go now

AlecAivazis (approved):
Nice! Thanks for putting this together. Mind opening up a copy of this onto 2.0?

PR #1478, opened 2025-04-23
[houdini-svelte] properly warn against using page/layout queries when not include-ing .js files in houdini.config.js

### LINKED ISSUE #1416: Console error after setting the `include` config to `src/**/*.{graphql,gql}`
### Describe the bug

We set the `include` config in the `houdini.config.js` to `src/**/*.{graphql,gql}` so that Houdini doesn't watch all JS/TS files all the time (which led to a full reload of the page as soon anything changed). But now we're getting the following error in the console even if we don't have queries in our `+page` or `+layout` files:

```
⚠️ You are using at least one page/layout query but aren't "include"ing .js files in your config. \n This will cause houdini to not pick up the generated files and prevent it from working as expected.
```

Could it be that the condition for outputting this warning is not correctly defined:

https://github.com/HoudiniGraphql/houdini/blob/c307841fc0994be1b18932dd1eb5c505a8a1fe8e/packages/houdini-svelte/src/plugin/codegen/routes/index.ts#L71-L79

In my opinion the condition should look like this:

```JavaScript
if (
	// Include path doesn't include js
	!config.include.some((path) => path.includes('js')) &&
	// And the plugin isn't configured to run in static mode
	(!plugin_config(config).static &&
	    // User uses layout/page queries
	    (uniquePageQueries.length > 0 ||
	    uniqueLayoutQueries.length > 0))
) {
```

### Reproduction

_No response_

rgon:
Fixes #1416

### DIFF
--- a/.changeset/purple-readers-punch.md
+++ b/.changeset/purple-readers-punch.md
@@ -1,0 +1,5 @@
+---
+'houdini-svelte': patch
+---
+
+properly warn against using page/layout queries when not include-ing .js files in houdini.config.js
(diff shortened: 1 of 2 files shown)


AlecAivazis (approved):
Thanks for putting this together!

PR #1485, opened 2025-04-25
Feat config writepulledschema

### LINKED ISSUE #1483: Allow read only schema paths
### Describe the feature

Some expected houdini use cases provide a read only graphql schema file. For example, a shared volume between containerized frontends & a backend that exports this schema. In this use case, you wouldn't want multiple houdini instances to be updating the same schema file.

Also file permissions could prevent writing this schema file in annoying ways. 

The pullSchema method has that option exported, but it is never utilized:
```ts
pullSchema(
	url: string,
	fetchTimeout: number,
	schemaPath: string,
	headers?: Record<string, string>,
	skipWriting?: boolean       // <----------
): Promise<string | null> {
```

## Proposal:

+ [ ] Gracefully error out of writing the schema file: file write does not affect houdini, and thus shouldn't prevent the codegen pipeline from running successfully.
+ [ ] Provide a way to watch the schema but not write it to a file:
```js
watchSchema: {
    url: ...
    skipWriting: true
}
```

IMO the default behaviour of writing the schema to a file should be kept, since it serves for IDE integration.

### Criticality

rgon:
Fixes #1483

Implement schema poll without writing to disk.
Gracefully error out of schema write, since that's not technically required for houdini codegen.

By extension, allows glob local schemaPath & api poll (wathSchema) at the same time.

### DIFF
--- a/packages/houdini/src/cmd/init.ts
+++ b/packages/houdini/src/cmd/init.ts
@@ -157,7 +157,7 @@ export async function init(
 				fetchTimeout,
 				schemaPath,
 				local_headers,
-				false
+				true
 			)
 
 			if (pullSchema_content === null) {
@@ -196,11 +196,6 @@ export async function init(
 		)
 
 		schemaPath = answers.schema_path
-	}
-
-	// Let's write the schema only now (after the function "after_questions" where the project has been created)
-	if (is_remote_endpoint && pullSchema_content) {
-		await fs.writeFile(path.join(targetPath, schemaPath), pullSchema_content)
 	}
 
 	// try to detect which tools they are using
--- a/packages/houdini/src/cmd/init.ts
+++ b/packages/houdini/src/cmd/init.ts
@@ -157,7 +157,7 @@ export async function init(
 				fetchTimeout,
 				schemaPath,
 				local_headers,
-				true
+				false
 			)
 
 			if (pullSchema_content === null) {
--- a/packages/houdini/src/lib/config.ts
+++ b/packages/houdini/src/lib/config.ts
@@ -59,6 +59,7 @@ export class Config {
 	routesDir: string
 	schemaPollInterval: number | null
 	schemaPollTimeout: number
+	schemaPollWriteToDisk: boolean = false
 	schemaPollHeaders:
 		| ((env: any) => Record<string, string>)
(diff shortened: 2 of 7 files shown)


packages/houdini/src/runtime/lib/config.ts
```diff
@@ -312,6 +312,13 @@ export type WatchSchemaConfig = {
 	headers?:
 		| Record<string, string | ((env: Record<string, string | undefined>) => string)>
 		| ((env: Record<string, string | undefined>) => Record<string, string>)
+
+	/**
+	 * Write the schema to disk on pull.
+	 * Useful for IDE integration.
+	 * Set to false when you have read only access to the schema or directory it's in.
+	 */
+	writePolledSchema?: boolean
```
SeppahBaws, line 321: Might be a good idea for clarification to mention here as well what its default behavior is in case it's not provided.

packages/houdini/src/lib/config.ts
```diff
@@ -163,6 +164,7 @@ export class Config {
 		this.routesDir = path.join(this.projectRoot, 'src', 'routes')
 		this.schemaPollInterval = watchSchema?.interval === undefined ? 2000 : watchSchema.interval
 		this.schemaPollTimeout = watchSchema?.timeout ?? 30000
+		this.schemaPollWriteToDisk = !(watchSchema?.writePolledSchema ?? true)
```
SeppahBaws, line 167: Is this logic correct, shouldn't it be without that `!` in the beginning?

packages/houdini/src/cmd/init.ts
```diff
@@ -157,7 +157,7 @@ export async function init(
 				fetchTimeout,
 				schemaPath,
 				local_headers,
-				true
+				false
```
SeppahBaws, line 160: Since this is inside the block for when we're using a remote schema, we should actually write the schema file when setting up a fresh project.

SeppahBaws (changes requested):

packages/houdini/src/lib/config.ts
rgon: Correct, I mesed up when removing the double negative from the option

SeppahBaws (approved):
Awesome, thank you for adding this!

PR #1491, opened 2025-05-22
[Houdini 2.0] Load event parameter types

SeppahBaws:
This is the same as #1487, but on the `houdini-2.0` branch.


AlecAivazis:
Ah nice! thanks

PR #1493, opened 2025-06-09
Constrain maximum version of kit

AlecAivazis:
This PR constrains the maximum version of valid kit versions when working with the current branch of Houdini. Unfortunately version 2.21.1 introduced a breaking change for the magic that powers `+page.gql` files. As much as it saddens me, I think we have to move away from that API in sveltekit projects for the foreseeable future until kit gives us a first-class ability to add routes dynmically. 

Also, the recent RFC for async support also seems to indicate that the Svelte team is moving in a direction more in line with NextJS. I think this is ultimately a good move for the ecosystem but it doesn't mesh well with the `+page.gql` style of loaders. In this new world, `houdini-svelte` will likely only be responsible for generating the stores and letting users decide how they want to invoke the store's load function. One very bright side to all of this is that we will become very decoupled from kit's capabilities and we can focus on what we know best - building a state of the art GraphQL client. A quick follow up to this PR will be another PR that introduces `houdini-svelte@3.0` as part of the `houdini-2.0` effort which removes the cap on the version along with all of the good bad and ugly necessary to support `+page.gql`.

### DIFF
--- a/site/package.json
+++ b/site/package.json
@@ -19,7 +19,7 @@
 	"devDependencies": {
 		"@babel/parser": "^7.20.7",
 		"@playwright/test": "1.30.0",
-		"@sveltejs/kit": "1.9.3",
+    "@sveltejs/kit": "<=2.21.0",
 		"@typescript-eslint/eslint-plugin": "^5.50.0",
 		"@typescript-eslint/parser": "^5.50.0",
 		"eslint": "^8.33.0",
--- a/site/package.json
+++ b/site/package.json
@@ -19,7 +19,7 @@
 	"devDependencies": {
 		"@babel/parser": "^7.20.7",
 		"@playwright/test": "1.30.0",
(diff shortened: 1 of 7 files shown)


site/package.json
```diff
@@ -19,7 +19,7 @@
 	"devDependencies": {
 		"@babel/parser": "^7.20.7",
 		"@playwright/test": "1.30.0",
-		"@sveltejs/kit": "1.9.3",
+    "@sveltejs/kit": "<=2.21.0",
```
SeppahBaws, line 22: There's some whitespace weirdness going on here

AlecAivazis: should be fixed now 👍

MaxPuckeridge:
so if we want to ignore this peer depenency and keep up to date with kit, should we stop using page.gql and use @load instead?

PR #1526, opened 2025-08-23
persistent query generation.

siddarthvader:
This PR brings support for generating persisted queries to compiler. Right now we dont care if the user passes `--output` flag or `persistedQueriesPath` is set in config so we generate the queries.json all the time.

For generating persisted queries we use a simple function that calls documents table and get all operations and make a named map of `fragments` and a id map of rest of the operations (`query, ,mutation, subscription`). Then we run another sql query against selection and selection_Refs table to find out what fragments are being used, then we get just them from the fragments map and prepare queries to append to a `json` file.

This PR also introduced one more fix that is we calculate and store `sha256` hash for documents in sqlite.

Tests suite is included and tests for cases like not having duplicate hash,invalid file extensions and such most of the test suite generate by claude code but tests cases were considerably added after actually facing some of the issues while writing code, for example earlier i was by mistake calcauting hash for entire documents but i ended up with duplicate hash in both `sqlite` and `queries.json`. Other tests just test basic things like file path/nested fragments and so on.


packages/houdini-core/plugin/documents/artifacts/collect.go
```diff
@@ -33,6 +34,7 @@ func CollectDocuments(
 		EnumValues:      map[string][]string{},
 	}
 
+	log.Println("Collecting documents")
```
AlecAivazis: Let's remove all of the print statements here. Eventually it would be good to figure out some kind of logging solution so that we can have these debugging logs when we need them.

packages/houdini-core/plugin/documents/artifacts/print.go
```diff
@@ -108,8 +111,15 @@ func printDocWorker(
 		// print the document
 		printed := PrintCollectedDocument(doc, includeHidden)
 
-		// update the document with the printed version
-		err = db.ExecStatement(update, map[string]any{"name": doc.Name, "printed": printed})
+		// generate hash from the printed content
+		hash := generateDocumentHash(printed)
+
+		// update the document with both printed version and hash
+		err = db.ExecStatement(update, map[string]any{
```
AlecAivazis, line 116: We need to use this value when we are generating the selection here: https://github.com/siddarthvader/houdini/blob/9a6b9aa18c99958155fcba5d95b7cf6056a89a97/packages/houdini-core/plugin/documents/artifacts/selection.go#L91

packages/houdini-core/plugin/documents/persistentQueries.go
```diff
@@ -0,0 +1,164 @@
...
+			return nil
+		},
+	})
+	if err != nil {
+		return plugins.WrapError(err)
+	}
+
+	for _, op := range operations {
+		// Get all fragment dependencies
+		fragmentNames := []string{}
```
AlecAivazis, line 84: Hm maybe i'm a bit confused but I'm not exactly following why we need to do the recursive query here. By the time the generation code is run, the collected documents struct should have the dependencies here: https://github.com/siddarthvader/houdini/blob/9a6b9aa18c99958155fcba5d95b7cf6056a89a97/packages/houdini-core/plugin/documents/artifacts/collect.go#L1179

packages/houdini-core/plugin/generate.go
```diff
@@ -3,16 +3,25 @@ package plugin
...
 
 func (p *HoudiniCore) Generate(ctx context.Context) error {
 	// the first thing to do is generate the artifacts
+
 	err := artifacts.Generate(ctx, p.DB, p.Fs, false)
 	if err != nil {
 		return err
 	}
 
+	defaultPersistentQueriesPath := "./queries.json"
```
AlecAivazis: Let's pull this value from the config file with a default. The default values get written here: https://github.com/siddarthvader/houdini/blob/9a6b9aa18c99958155fcba5d95b7cf6056a89a97/packages/houdini/src/lib/database.ts#L412.

You can pull the config value into the plugin using this method: https://github.com/siddarthvader/houdini/blob/9a6b9aa18c99958155fcba5d95b7cf6056a89a97/packages/houdini-core/plugin/documents/artifacts/generate.go#L44

AlecAivazis (changes requested):

packages/houdini/src/lib/project.ts
```diff
@@ -36,6 +36,7 @@ export const default_config: ConfigFile = {
 	defaultPaginateMode: 'Infinite',
   defaultFragmentMasking: 'enable',
   defaultCachePolicy: 'CacheOrNetwork',
+  persistedQueriesPath: 'queries.json',
```
AlecAivazis: this would put them in `<project_root>/queries.json` right? the old implementation put the file in `<runtime_dir>/queries.json`. Let's use the old value

AlecAivazis (changes requested):

packages/houdini-core/plugin/documents/artifacts/selection.go
```diff
@@ -410,22 +413,24 @@ func printedValue(
...
+		return d, err
 	}
 	defer query.Finalize()
 
 	err = db.StepStatement(ctx, query, func() {
-		printed += query.GetText("printed") + "\n\n"
+		if query.GetText("name") == name {
+			d.Hash = query.GetText("hash")
+		}
+		d.Printed += query.GetText("printed") + "\n\n"
```
AlecAivazis, line 425: Something feels wrong here - can you explain why we have to append Printed values but the hash is good to use? the hash should be the hash on the full document (including the dependencies)

siddarthvader: so I'm checking whichever query name is equal to the name (parent doc) and then using that hash. It's the hash of the parent document from db.

siddarthvader: you mean we got to make a new hash of the printed values rather than using the  hash that's stored in SQLite for the parent/original document?

AlecAivazis: the hash should be on the full document - if a dependent document changes, the hash needs to change so that the server sees it as a whole new value since the query it will receive will be different

AlecAivazis (approved):
Thanks for pushing this through!

PR #1534, opened 2025-09-17
Schema/enums and definitions generation support.

siddarthvader:
Now generating the `schema` files, `enums` and `definitions` for custom directive.

All tests are passing.

There is a test called "Writing twice to "Writing twice doesn't duplicate definitions", for this i dont run the whole pipeline(`runPipeline` fn*) two times, i only run the schema definition generatio.

`runPipeline` Fn does not fully represent the series of operation when i run "`houdini generate` and since we are only testing the compiler layer i only ran `schemaDefinition` twice to assume it would be sufficient condition, i may be wrong.


packages/houdini-core/plugin/schema/generateDefinitions.go
```diff
@@ -10,32 +13,349 @@ import (
...
+}
+
+func generateEnumFiles(ctx context.Context, db plugins.DatabasePool[config.PluginConfig], fs afero.Fs, projectConfig plugins.ProjectConfig) error {
+	var enumString strings.Builder
+
+
+	// just joining the tables and getting the exact js enum out of it using group_concat and sql templating
+	err := db.StepQuery(ctx, `
+		SELECT t.name,
+		       'export const ' || t.name || ' = {' || char(10) ||
```
AlecAivazis: It seems more complicated to do this inside of the query instead of building up the string in go. Let's refactor this so we build up the final definitions in a struct outside of the queries and then generate the files as a templated string

packages/houdini-core/plugin/schema/generateDefinitions.go
```diff
@@ -10,32 +13,349 @@ import (
...
+
+	// now we want to generate enums.d.ts file
+	var tsEnumString strings.Builder
+
+	// add the ValuesOf helper type at the top
+	tsEnumString.WriteString("type ValuesOf<T> = T[keyof T]\n\t\n")
+
+	err = db.StepQuery(ctx, `
+		SELECT t.name,
+		       'export declare const ' || t.name || ': {' || char(10) ||
```
AlecAivazis: same here as above

AlecAivazis (changes requested):

packages/houdini-core/plugin/schema/generateDefinitions.go
siddarthvader: updated.

packages/houdini-core/plugin/schema/generateDefinitions.go
siddarthvader: updated

packages/houdini-core/plugin/schema/generateDefinitions.go
```diff
@@ -10,32 +13,387 @@ import (
 	"code.houdinigraphql.com/plugins"
 )
 
+type Directive struct {
```
AlecAivazis: can we make these structs private (ie start with a lower case letter) since they aren't meant to be used by anything outside of these functions. If posisble we could also just move their definition inside of the generator

siddarthvader: updated.

siddarthvader: moved the definition inside generator function, removed extrenal js/ts enum genration function, all logic inline now.

PR #1538, opened 2025-09-30
Schema/enums and definitions generation support.

siddarthvader:
Now generating the schema files, enums and definitions for custom directive.

All tests are passing.

There is a test called "Writing twice to "Writing twice doesn't duplicate definitions", for this i dont run the whole pipeline(runPipeline fn*) two times, i only run the schema definition generatio.

runPipeline Fn does not fully represent the series of operation when i run "houdini generate and since we are only testing the compiler layer i only ran schemaDefinition twice to assume it would be sufficient condition, i may be wrong.

### DIFF
--- a/packages/houdini-core/plugin/schema/generateDefinitions_test.go
+++ b/packages/houdini-core/plugin/schema/generateDefinitions_test.go
@@ -15,6 +15,259 @@ import (
 	"code.houdinigraphql.com/plugins/tests"
 )
 
+func TestDefinitionGeneration(t *testing.T) {
+	tests.RunTable(t, tests.Table[config.PluginConfig]{
+		Tests: []tests.Test[config.PluginConfig]{
+			{
+				Name: "Generates runtime definitions for each enum",
+				Input: []string{
+					`query GetAppVersion { version }`,
+				},
+				Extra: map[string]any{
+					"enumsTypesContains": []string{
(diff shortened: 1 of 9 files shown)


packages/houdini-core/plugin/schema/generateDefinitions_test.go
```diff
@@ -0,0 +1,429 @@
...
+				Value3
+				Value2
+			}
+    `,
+		PerformTest: func(t *testing.T, p *plugin.HoudiniCore, test tests.Test[config.PluginConfig]) {
+			config, err := p.DB.ProjectConfig(context.Background())
+			assert.Nil(t, err)
+
+			switch test.Name {
+			case "Generates runtime definitions for each enum":
```
AlecAivazis, line 149: Let's use the Extra field in the test in order to keep each test case isolated. you can save the expected values in the map and then look up the value in `test.Extra` inside of `PerformTest`

AlecAivazis:
When I run the generate command in the `e2e/react` project I see a bunch of empty objects:

image


The old tests validated entire snapshots of the files - let's keep that structure so we can make sure that we perfectly reproduce the previous logic

AlecAivazis (approved):
Looks great! Thanks for fixing the tests 👍

PR #1549, opened 2025-10-17
websocket implementation

siddarthvader:
implemneted websockets networking layer along with HTTP for hooks invocation rather we have bidirectional communication now between compiler and orchestrator. There is no schema change, DB already has a column for port in plugin table.

 
**Since we have plugin to plugin hook invocation also, we are using our existing http layer for that, we have abstracted the hook handlers into a new files called handlers.go and calling these functions form both http and websocket layer.**

We maintain a map of websocket connection, and we keep an account of pending request, this is still request-response model nothing inherent has changed just the networking layer is websocket now.

This improved upon the existing http networking, now our golang process can send message to orchestrator (posing as websocket client) there is a simple interface for messages implemented.

Right now we dont send a lot of info from compiler using the websockets, only sending critical errors propagating in hook invocation and their success message.

Earlier we sent `taskId` and `pluginDir` as headers

```
{
  method: 'POST',
  headers: {
	  'X-Task-ID': task_id?.toString() ?? '',
	  'X-Plugin-Directory': directory,
  },
  body: JSON.stringify(payload),
},
```

```
type WebSocketMessage struct {
	ID              string         `json:"id"`
	Type            string         `json:"type"`
	Hook            string         `json:"hook"`
	Payload         map[string]any `json:"payload"`
	TaskID          string         `json:"taskId"`
	PluginDirectory string         `json:"pluginDirectory"`
}

```

Now we can just send them as regular payload.

We have a lot of errors already propagated to hook layer and then being sent to client, looks somewhat liks this 
image


packages/houdini/src/lib/codegen.ts
```diff
@@ -217,36 +220,125 @@ export async function codegen_setup(
...
+	async function getOrCreateWS(name: string, port: number): Promise<WebSocket> {
+		const existing = wsConnections.get(name)
+		if (existing && existing.readyState === WebSocket.OPEN) {
+			return existing
 		}
-		return await response.text()
+
+		return new Promise((resolve, reject) => {
+			// get a new connection
+			const wsUrl = `ws://localhost:${port}/ws`
```
AlecAivazis, line 232: Let's make the root of the plugin web server be the websocket endpoint

packages/houdini/src/lib/codegen.ts
```diff
@@ -217,36 +220,125 @@ export async function codegen_setup(
...
+		hook: string,
+		payload: Record<string, any> = {},
+		task_id: string | undefined,
+		port: number,
+		directory: string
+	): Promise<any> => {
+		const ws = await getOrCreateWS(name, port)
+
+		return new Promise((resolve, reject) => {
+			const messageId = `${hook}-${Date.now()}-${randomUUID()}`
```
AlecAivazis: Since there is only ever one orchestrator for a given plugin (one source of messages), I think we can simplify this a bit and define a `messageID` variable in global scope that we increment on every message.

packages/houdini/src/lib/codegen.ts
```diff
@@ -217,36 +220,125 @@ export async function codegen_setup(
...
+
+			ws.on('error', (err) => {
+				console.error(`WebSocket error for ${name}:`, err)
+				wsConnections.delete(name)
+				reject(new Error(`WebSocket error for ${name}: ${err}`))
+			})
+
+			ws.on('close', () => {
+				// Remove from pool so next request creates new connection
+				// requestsd will eventually timeout and be rejected, we can agressively remove them but it's not a big deal
```
AlecAivazis:
```suggestion
				// requests will eventually timeout and be rejected, we can agressively remove them but it's not a big deal
```

plugins/context.go
```diff
@@ -28,6 +34,35 @@ func PluginDirFromContext(ctx context.Context) string {
...
-type pluginDirCtxKey struct{}
+// example of how to use the WSConnFromContext function
+//if conn := WSConnFromContext(ctx); conn != nil {
+// 	conn.WriteJSON(WebSocketResponse{
+// 		ID:    ctx.Value("wsMessageID").(string),
+// 		Type:  "1error",
+// 		Error: "TEST: This is a simulated non-fatal error during generation",
+// 	})
+// }
+// helper to get the conn from context,
```
AlecAivazis:
```suggestion
// helper to get the conn from context
//
// Example:
//  	if conn := WSConnFromContext(ctx); conn != nil {
// 			conn.WriteJSON(WebSocketResponse{
// 				ID:    ctx.Value("wsMessageID").(string),
// 				Type:  "1error",
// 				Error: "TEST: This is a simulated non-fatal error during generation",
// 			})
// 		}
```

plugins/context.go
```diff
@@ -28,6 +34,35 @@ func PluginDirFromContext(ctx context.Context) string {
 	return ctx.Value(pluginDirCtxKey{}).(string)
 }
 
-type taskIDCtxKey struct{}
+func ContextWithWSConn(ctx context.Context, conn *websocket.Conn) context.Context {
+	return context.WithValue(ctx, "wsConn", conn)
```
AlecAivazis: let's use a non-string key for the context here. Like you did for `taskIDCtxKey`

plugins/run.go
```diff
@@ -55,12 +56,37 @@ func Run[PluginConfig any](plugin HoudiniPlugin[PluginConfig]) error {
...
+		// upgrade to websocket
+		conn, err := upgrader.Upgrade(w, r, nil)
+		if err != nil {
+			log.Fatal(err)
+			return
+		}
+		defer conn.Close()
+
+		// connection established
+		log.Printf("ws connection established from %s", r.RemoteAddr)
```
AlecAivazis: This comment reminds me that we need to have a good story for different levels of logs. This kind of message is really helpful for debugging but something we probably don't want to show the user.

plugins/websocket.go
```diff
@@ -0,0 +1,471 @@
...
+		}
+	}
+}
+
+// generator plugin functions
+func handleGenerateRuntime[PluginConfig any](
+	plugin HoudiniPlugin[PluginConfig],
+) func(ctx context.Context, payload map[string]any) (any, error) {
+	return func(ctx context.Context, payload map[string]any) (any, error) {
+		// if conn := WSConnFromContext(ctx); conn != nil {
```
AlecAivazis: let's remove this comment

plugins/websocket.go
```diff
@@ -0,0 +1,471 @@
...
+func pluginWebsocketHooks(plugin HoudiniPlugin[config.PluginConfig]) []string {
+	hooks := []string{}
+
+	// --- Config
+	hooks = append(hooks, "Config")
+	registerWSHandler("Config", handleConfig(plugin))
+
+	// --- AfterLoad is triggered for StaticRuntime OR AfterLoad
+	_, isStaticRuntime := plugin.(StaticRuntime)
+	_, isAfterLoad := plugin.(AfterLoad)
```
AlecAivazis: This file is missing some of the changes to the hook logic that was added as part of the svelte plugin PR. The one that stood out to me here was the `DefaultConfig` interface but there could be others

AlecAivazis (changes requested):
Overall this looks really good! It's a significant upgrade from the current setup so thanks for pushing on this.

plugins/run.go
siddarthvader: yes we could expose interface with log-levels, most def something we can implement. will leave a TODO comment.

plugins/websocket.go
siddarthvader: ported the remaining ones from http.go

plugins/context.go
siddarthvader: done

plugins/websocket.go
siddarthvader: done

packages/houdini/src/lib/codegen.ts
siddarthvader: done

plugins/context.go
siddarthvader: done

siddarthvader:
@AlecAivazis fixes pushed for all comments, kindly review.

packages/houdini/src/lib/codegen.ts
```diff
@@ -219,36 +221,127 @@ export async function codegen_setup(
...
-			method: 'POST',
-			headers: {
-				'Content-Type': 'application/json',
-				'X-Task-ID': task_id?.toString() ?? '',
-				'X-Plugin-Directory': directory,
-			},
-			body: JSON.stringify(payload),
-		})
+		// All hooks now use WebSocket
+		return await invoke_hook_websocket(name, hook, payload, task_id, port, directory)
```
AlecAivazis: let's remove the `_websocket` and stick with just `invoke_hook` since there is only a websocket version now

AlecAivazis (changes requested):

packages/houdini/src/lib/codegen.ts
siddarthvader: done

AlecAivazis (approved):
Thanks for getting this together! I think its finally good to go 🚀

PR #1723, opened 2026-07-09
session refresh queries

### LINKED ISSUE #1225: Changing session should refresh queries
### Describe the feature

Some improvements should be made to the session handling in houdini, in particular the ability to invalidate session dependent queries would be cool. This may or may not be blocked by #891. The thread also discusses some "extra" stuff like being able to absorb an existing response payload and update the cache without the request, but that is likely out of scope for a single ticket.

Previously discussed here: https://discord.com/channels/1024421016405016718/1164588891605565440

Some quotes to try and summarize the issue, I'll clean this up later if I have time:

> yea that's what I'm thinking too. since the logic _has_ to be on the server, the simplest API is just to provide some function you call on the server to update the session. `setSession` is the best API - it doesn't exactly match the existing use but maybe it's okay having the one just because its easy to remember cc <@198142399002836992> 
> 
> i think i'm going to wire it up to refresh all of the active queries (maybe we could explore some directive like `@sessionless` to say that a query doesn't depend on the session). I think it's the easiest to understand and it sets us up for being able to do something like to rerefresh a single entry in the cache
> 
> ```
> const user = cache.get('User', { id: "1"})
> 
> user.refresh()
> ```

> yea re-running the queries will take some plumbing. the current channel between the cache and stores is just used to send data but now the cache needs to be able to "talk" to the active stores and send signals that they can respond to do to things

> it's not gonna be that tricky - just gotta nest the current payload under some `kind: 'update'` and it opens up all sorts of neat things if we have more kinds of messages

> could that be a signal sent by the server? so setSession does something to event.locals which causes a reload?
> > no i'm thinking that's a separate part
> > something on the client will need to detect the changd state and retrigger the queries
we already do the detection to keep the client-side singleton in sync

> i wonder if you could sort of pack this change into having houdini generally play nice with form actions, like if you could run a mutation on the server and plumb the result through to the client to update the cache if enhanced
>>yea its the must-have escape hatch before we get into the fine-grained stuff
>
>>ultimately I'd like to have a way for you to push abitrary updates from the server to the client and have it update the cache (ie, you just need to send the mutation payload back and move on)
>
>>the biggest hurdle is that features like these need to take a lot of opinions that could conflict with people's setups. ie if we push the events from the server to the client via websockets then spa adapters don't work. so the trick is finding the right way to package it up so that users naturally fall into a pit of success but that's tricky
>
>>so until we figure that bit out, we can just worry on the refreshing bit

### Criticality

siddarthvader:
Fixes #1225 

Changes in houdini session will trigger all queries to refetch. When `__houdini_session__` is changed because of external factors or anyhow, we already watch page.data for setting client session but now we also deep compare the old `_houdini__session_` with latest one,if it has changes we will refresh all the queries with new session

The cache refetch message can now carry a session value. Query refetches use that value when present instead of falling back to the query's last session, which prevents session-change refreshes from refetching with

```
export type CacheMessage<_Data = any> =
	| {
			kind: 'update'
			data: _Data
	  }
	| {
			kind: 'refetch'
			session?: App.Session | null // added new
	  }

```

 

packages/houdini/src/runtime/cache/subscription.ts
```diff
@@ -348,6 +348,16 @@ export class InMemorySubscriptions {
 		)
 	}
 
+	all({ includeMaskedParents = false }: { includeMaskedParents?: boolean } = {}) {
+		return [...this.subscribers.values()].flatMap((fields) =>
```
AlecAivazis: Let's convert this to nested `for ... in` so that we don't pay the price of walking through `fields.values` multiple times. We can just build up the unique selections that are relevant without having to then build up `notified` in the other call site. `all` shoudl just return the unique list

AlecAivazis (changes requested):

packages/houdini/src/runtime/cache/subscription.ts
siddarthvader: resolved.

packages/houdini/src/runtime/types.ts
```diff
@@ -319,6 +319,7 @@ export type CacheMessage<_Data = any> =
 	  }
 	| {
 			kind: 'refetch'
+			session?: App.Session | null
```
AlecAivazis, line 322: we should also add metadata and the other extra args from the fetch functino

AlecAivazis (changes requested):

packages/houdini-svelte/package/vite/transform/session.test.ts
```diff
@@ -14,15 +14,25 @@ test('modifies root +layout.svelte with data prop', async function () {
 	)
 
 	expect(result).toMatchInlineSnapshot(`
```
AlecAivazis, line 16: Let's add an e2e test for this logic

AlecAivazis (changes requested):

packages/houdini-svelte/package/vite/transform/session.test.ts
siddarthvader: done

AlecAivazis (approved):

PR #1746, opened 2026-09-03
fix: Windows plugin process spawn

knd775:
Fixes ["Windows launcher issue"](https://discord.com/channels/1024421016405016718/1528401921193541692) thread on discord

The houdini compiler would crash on Windows when it attempted to load plugins.


AlecAivazis (approved):
Thanks for getting this out!

PR #1751, opened 2026-09-14
fix: Runtime scalars failing validation during HMR and builds

knd775:
Runtime scalar variable ids were being saved in the `argument_values(document)` column, which is a FK for `documents(id)`. By coincidence, this worked most of the time becuase there are typically more documents than runtime scalar variables. However, during HMR the ids were autoincremented, which inevitibly lead to a FK violation when the variable ids didn't happen to match a document id anymore.

Now, the `document` column correctly holds a document id and FK violations aren't silently dropped anymore.


knd775:
I should probably note that this doesn't only affect HMR. It has a chance to break any client generation. In our application, this means about 10% of production builds fail.

